Issue #81, opened 2025-09-03
Contribute


verytactical:
Documentation and ecosystem contributions with focus on the former.

novusnota:
Let's change the scope and move it to the bottom of the side navigation.

UPD: done.

PR #181, opened 2025-09-11
create contribution section, move all pages from the older structure to /pending

### LINKED ISSUE #180: Move folders and pages from the older structure to /pending
Whilst updating the redirects.

novusnota:
Closes #180
Towards #81


novusnota:
This is just so we don't confuse the structures while making sure no pages are lost until we're done with them. At that point, we'll remove the `/pending` folder entirely.

Issue #141, opened 2025-09-09
Fine-tune the Vale's settings and dictionary, and add it to package.json's scripts


anton-trunov:
<img width="842" height="456" alt="Image" src="https://github.com/user-attachments/assets/7457fb0d-77db-474e-bb02-b45b8f036bce" />

https://www.mintlify.com/docs/settings/ci#configuration-2

PR #187, opened 2025-09-11
Fill the Account states page

### LINKED ISSUE #170: [Foundations > Account status]
Old articles
- https://docs.ton.org/v3/documentation/smart-contracts/addresses/address-states

Karkarmath:
Closes #170

### DIFF
--- a/.vale/config/vocabularies/Custom/accept.txt
+++ b/.vale/config/vocabularies/Custom/accept.txt
@@ -92,8 +92,6 @@ Webserver
 Whitepapers
 Workchain
 Yolo
-account_state_changed
-action_phase_is_successful
 adnl
 allowlist
 altcoins
@@ -200,7 +198,6 @@ sbt
 scannability
 sdks
 semver
-send_dest_if_zero
 seqno
 sharded
 sharding
@@ -208,7 +205,6 @@ skillset
 sse
 stablecoin
 standart
-storage_fees_due
 styleguide
 subnetworks
 subslice
@@ -239,4 +235,3 @@ whitepaper
 whitepapers
 workchain
 workchains
-zero_bal_after_dest_act
--- a/ton/states.mdx
+++ b/ton/states.mdx
@@ -2,9 +2,6 @@
 title: "Account statuses"
 ---
 
-
-[//]: # (import ZoomableImage from '@site/src/components/ZoomableImage')
-
 This article describes the four possible states of an account on TON Blockchain.
 Understanding these states is crucial for accurately predicting transaction outcomes and ensuring the correct deployment.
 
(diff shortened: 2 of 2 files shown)


ton/states.mdx
```diff
@@ -1,5 +1,134 @@
 ---
-title: "Contract states"
+title: "Address statuses"
```
verytactical: Are they statuses or states? :)

This is not consistent to the text below

ton/states.mdx
```diff
@@ -1,5 +1,134 @@
...
 ---
 
-- https://docs.ton.org/v3/documentation/smart-contracts/addresses/address-states
+[//]: # (import ZoomableImage from '@site/src/components/ZoomableImage')
```
verytactical: Also could you please check whether the image has to go into `resources/images/`, or we can put it next to the mdx file?

It's generally not a good idea to keep related files far apart over filesystem, and if Mintlify doesn't support that, we'd better raise it as an issue.

novusnota: Mintlify does support putting an image next to the mdx file, or anywhere else, really. The `/resources/images/` folder is just a product of my actions, to ensure clean-looking links.

That said, if an image is to be used in multiple places, it's better to place it in `/resources/images` :)

ton/states.mdx
novusnota: Perhaps, it should be titled: "Account and address states"? Former is the actual term for this (account states), while the latter is common slang.

Alternatively, we can name this page "Accounts", place before "Addresses", and also explain what accounts are in addition to their states/statuses descriptions given here already.

Karkarmath: Well, according to tl-b they are AccountStatuses. But we seemed to be discussing what is more correct to say address status, since almost always an account means an address that already has a code and date. And the Account state is often understood as its state_init.

Karkarmath: Anton and I talked about this topic and decided to end up with Account Status, as in tlb. I'll fix everything here soon.

ton/states.mdx
verytactical: Created #240 to handle this. Let's ignore this for now.

ton/states.mdx
```diff
@@ -1,5 +1,134 @@
...
+
+## What is the account status?
+
+The account status is a formal indicator of what actions can occur with an account, what it can store, whether it contains a contract code or not. In other words, it is one of the main factors determining the behavior of a given account during a transaction. The account status at the beginning and end of a transaction is recorded in [the corresponding TL-B block](https://github.com/ton-blockchain/ton/blob/cac968f77dfa5a14e63db40190bda549f0eaf746/crypto/block/block.tlb#L285) in the fields `orig_status` and `end_status`. Thus, it allows developers to always view the current status of an account before sending it a message and restore the history of its status changes.  
+
+## Status variety
+
+Each account exists in one of the following statuses:
+
+-   **nonexist**: the default status for accounts with no transaction history or that were deleted. Contains no code, data, or balance. All 2<sup>256</sup> accounts start in this status.
```
verytactical, line 16: We still have a lot of mintlify failures even though #141 was fixed.

I think this PR should be rebased, and all the words added to [custom word list](https://github.com/tact-lang/mintlify-ton-docs/blob/c6585d8c44ee8ba0f6c162e177802a135307d64b/.vale/config/vocabularies/Custom/accept.txt).

ton/states.mdx
```diff
@@ -132,3 +133,4 @@ With the diagram **Legend** below, you can inspect all possible changes in the a
 - Each new `state_init` is ignored when the account status is active.
 
 <Feedback />
```
novusnota:
```suggestion
```

ton/states.mdx
```diff
@@ -1,5 +1,136 @@
 ---
-title: "Contract states"
+title: "Account statuses"
 ---
 
-Stub
+
+[//]: # (import ZoomableImage from '@site/src/components/ZoomableImage')
```
novusnota:
```suggestion
```

.vale/config/vocabularies/Custom/accept.txt
```diff
@@ -234,3 +239,4 @@ whitepaper
 whitepapers
 workchain
 workchains
+zero_bal_after_dest_act
```
novusnota: Hmm, let's not add exact code names here, and instead wrap those in backticks `` in text

Karkarmath: Well, ok.

ton/states.mdx
Karkarmath: done

ton/states.mdx
Karkarmath: done

PR #195, opened 2025-09-15
`<Image>` component

### LINKED ISSUE #190: Make an `<Image>` React component
We need:
- To present SVGs with ability to zoom, pan or pinch them
- To present JPG and PNG images for light and dark themes together
  - Furthermore, we might invert colors in SVGs to achieve that effect with a single file
- To have a basic zoom option for raster images

Currently, Mintlify only provides us with the last option. However, we need more power to our images, which would help cases like #187 and many more.

novusnota:
Closes #190.

The `<Image>` component:
1. has the featureset of the `ThemedImage` from the current TON Docs, but with less hassle — specify `src` and `darkSrc` paths and you're done with a dual-theme image showcase.
2. inverts colors if an SVG is provided in the `src` and there is no `darkSrc` alternative.
3. (needs debugging) if an SVG is provided and `svgControls` property is set, shows the SVG on a neat zoom/pan/pinch pane, with corresponding controls. At least that's what is planned and ported from the `ZoomableImage` component in the current TON Docs.

In any case, even the first two points are enough to make #187 somewhat better.

Now, regarding the 3rd point: no matter which approach I took, I couldn't make it work in Mintlify. I might be missing something, hence the @verytactical tag :)

We can ask the Mintlify team to prepare some registry for us — that is, to allow writing `import {...} from "name-of-the-lib-we-asked-about";`. Then, our React components could import stuff from npm in addition to the basic `"react"` imports.

### DIFF
--- a/.vale.ini
+++ b/.vale.ini
@@ -43,14 +43,6 @@ BlockIgnores = (?sm)^(<\w+\n .*\s\/>)$, \
 # {/* comments */}
 CommentDelimiters = {/*, */}
 
-# Ignoring the /pending folder
-[pending/*.{md,mdx}]
+# Completely ignore the /pending folder
+[pending/*]
 BasedOnStyles = NONE
-[pending/**/*.{md,mdx}]
-BasedOnStyles = NONE
-
-# [pending/blockchain-of-blockchains.mdx]
-# BasedOnStyles =
-# Vale = NO
-# [pending/*.{md,mdx}]
-# BasedOnStyles =


novusnota:
Notes to self:
1. Using prior approaches without an explicit `import` (importmaps or regular script tags), try importing `useRef` ([and others](https://cdn.jsdelivr.net/npm/react-zoom-pan-pinch@3.7.0/dist/index.esm.js)) and see if that will fix issues.
2. Try downloading [`react-zoom-pan-pinch`](https://www.npmjs.com/package/react-zoom-pan-pinch) locally ([regular](https://cdn.jsdelivr.net/npm/react-zoom-pan-pinch@3.7.0/dist/index.esm.js), [minified](https://cdn.jsdelivr.net/npm/react-zoom-pan-pinch@3.7.0/dist/index.esm.min.js)), and then using it from the `/resources` folder.

If that fails, just request support of `react-zoom-pan-pinch` from the Mintlify team in their registry as discussed prior in early August meetings. It should be very doable, and I suspect some other users of Mintlify's Enterprise/Business/Custom tier do it already.

UPD: Perhaps, we also need custom support for [tanstack table libs](https://tanstack.com/table/latest), `@ton/core`, `assets-sdk`, `@ton/ton`, and `@tonconnect/ui-react`.

verytactical:
Please raise the issue with Mintlify team directly

novusnota:
Couple of points:
1. I've removed the attempts to add a third-party npm package for this component. It should be noted that packages with UMD module resolution would work in Mintlify quite easy, but `react-zoom-pin-pinch` has ESM-only setup, so that fails.
2. I've added an ability for specifying a `href` and a `target` to allow making images into clickable links.
3. I've tried fixing Mintlify's Vale CI issues, but I could only resolve those locally — Mintlify doesn't seem to be picking up path ignore rules used in Vale's `.vale.ini` config. That can be fixed by disabling spell checks provided by Mintlify and adding our own GitHub Action for those. This might be a temporary solution to #312.

Anyways, let's merge this `<Image>` component, and I'll write some docs on its usage as per #272 shortly after.

PR #246, opened 2025-09-16
veni, vidi, nominavi, movi

novusnota:
As discussed with @verytactical. Also, decided to keep "Web3 services" where they are since those aren't something front and center afaict. Addred comments.

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -38,7 +38,7 @@
       "guidebook/start-here",
       "guidebook/get-support",
       {
-        "group": "Step-by-step guides",
+        "group": "Step by step",
         "pages": [
           "guidebook/first-dapp",
           "guidebook/payment",


docs.json
```diff
@@ -54,6 +54,19 @@
           "guidebook/dapp",
           "guidebook/wallet",
           "guidebook/cex",
+          {
```
novusnota, line 57: That's just in case, but those can be removed and introduced again in the future

docs.json
```diff
@@ -38,7 +38,7 @@
       "guidebook/start-here",
       "guidebook/get-support",
       {
-        "group": "Step by step",
+        "group": "Step-by-step guides",
```
verytactical: Not catchy anymore. I'd rather not rename this

docs.json
```diff
@@ -144,6 +157,45 @@
           "ecosystem/more"
         ]
       },
+      {
```
novusnota, line 160: 1. step-by-step
2. ecosystem
3. standard contracts
4. contract patterns
5. languages
6. tvm ("where and how language outputs, compiled BoCs, are executed")
7. blockchain foundations
8. web3 services — aren't particularly fancy right now, but we might move 'em closer to ecosystem if such need arises

docs.json
```diff
@@ -144,6 +157,45 @@
...
+              "standard/tokens/airdrop"
+            ]
+          },
+          "standard/dex",
+          "standard/vesting",
+          "standard/governance"
+        ]
+      },
+      {
+        "group": "Contract patterns",
```
verytactical, line 186: "Smart-contract" was intentionally there too :)

docs.json
novusnota:
```suggestion
        "group": "Step by step",
```

novusnota: ok, restored

docs.json
novusnota: Hmm, both sections start with "S", and that made me misread it once during our sync, hence this change

verytactical: Wasn't my idea, and I didn't really like it in the first place, so...

Let's keep "contracts"

PR #316, opened 2025-09-19
twoslash test

novusnota:
It's a test to see whether Mintlify's code blocks truly support the [Twoslash engine](https://twoslash.netlify.app)

### DIFF
--- a/start-here.mdx
+++ b/start-here.mdx
@@ -3,3 +3,12 @@ title: "Start here"
 ---
 
 Stub
+
+```ts
+// @errors: 2339
+let x: [string, number]
+x = ['hello', 10] // OK
+/// ---cut---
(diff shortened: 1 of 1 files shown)


novusnota:
> It's a test to see whether Mintlify's code blocks truly support the [Twoslash engine](https://twoslash.netlify.app/)

Nope, they do not ([just yet](https://github.com/mintlify/mdx/commits/main/)). This might be a case only for the staging, because the search/AI there are disabled, and are only present for the main deployment. However, I don't think that's true, since Twoslash is part of the Markdown transformation pipeline, and if it's not enabled for staging, it's not enabled elsewhere.

Let's wait a bit for the official release and a feature trigger in the Mintlify dashboard settings:

image

PR #325, opened 2025-09-22
convert ton pdf to markdown

fakela:


fakela:
Mintlify supports latex for rendering mathematical expressions and equations which was needed in the docs . i used the 

` ```maths fence`

 for some code blocks. https://www.mintlify.com/docs/text#mathematical-expressions
 
 - Block Sequence Number  does not appear twice
 - workchain_id and co are in backticks

PR #354, opened 2025-09-22
add glossary

### LINKED ISSUE #230: [Foundations > Glossary]
Temporarily move it here, until we make wiki-like cross-references work

Should be at the end of the list

https://docs.ton.org/v3/concepts/glossary

verytactical:
Mostly just moves old page.

Minor fixes:
- remove terms not really suitable for documentation (`YOLO`);
- consistently have term and abbreviation in front of dash sign.

Closes #230

### DIFF
--- a/ton/glossary.mdx
+++ b/ton/glossary.mdx
@@ -42,7 +42,7 @@ a distributed ledger of data in the form of a chain of blocks recording transact
 
 ### Bag of Cells (BoC)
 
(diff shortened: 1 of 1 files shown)


reveloper:
@verytactical 
Can we keep the style of casing for words:
WorkChain, ShardChain, BaseChain, AccountChain, according to the legacy documentation?

verytactical:
There is an official way these words are written.

image

PR #374, opened 2025-09-22
feat(external-message): Hash normalization and message lookup (copypasted from TonConnect docs)

### LINKED ISSUE #44: [Foundations > Messages & transactions]
On the wire:

- https://docs.ton.org/v3/documentation/data-formats/tlb/msg-tlb
- https://docs.ton.org/v3/documentation/data-formats/tlb/block-layout
- https://docs.ton.org/v3/documentation/data-formats/tlb/transaction-layout
- https://docs.ton.org/v3/documentation/data-formats/tlb/crc32

User feedback 
- concept of a bouncable message is confusing

Shvandre:
Closes #44

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -167,6 +167,7 @@
             "group": "RPC providers",
             "pages": [
               "ecosystem/rpc/overview",
+              "ecosystem/rpc/external-normalization",
               "ecosystem/rpc/toncenter"
             ]
           },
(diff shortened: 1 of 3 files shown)


anton-trunov:
We need a few adjustements here:
- rename the article to better explain its contents, it's not really about normalization;
- clearly explain we are talking about looking for external messages from TON Connect to use for the purposes of UI, not payment processing;
- move to TON Connect section;
- make it clear TEP-467 is not a real TEP, but a TEP proposal that became a standard de-facto.

Issue #171, opened 2025-09-11
[Foundations] Consensus


verytactical:
- Rename "Block generation" to "Consensus"
- https://companyname-a7d5b98e.mintlify.app/ton/blocks
- catchain.pdf
- https://docs.ton.dev/86757ecb2/p/07ddda-walk-through-the-catchain

reveloper:
upd: added usefull source link https://docs.ton.dev/86757ecb2/p/07ddda-walk-through-the-catchain

PR #475, opened 2025-09-26
add TON consensus doc

laviniat1996:
This PR adds an indepth doc about how consensus works in TON. The doc still needs improvements, but it is in good shape to be added as it is.

Closes: https://github.com/tact-lang/mintlify-ton-docs/issues/171


anton-trunov:
@laviniat1996 please add "closes #..." to your original post, so we know which task this PR is solving

reveloper:
Added link to issue

ton/consensus.mdx
```diff
@@ -1,5 +1,225 @@
...
+  - A validator’s efficiency = `signedBlocks / expectedBlocks`.  
+  - If efficiency is lower than 90% in a round, fines are applied.  
+
+- **Liveness bound:**  
+  - Consensus always finalizes within a bounded number of attempts.  
+  - Null blocks prevent deadlock even under heavy network partitions.
+
+---
+
+## Summary
```
verytactical, line 217: Why does it have a summary?

verytactical (changes requested):
This seems to be AI generated.

ton/consensus.mdx
```diff
@@ -1,5 +1,225 @@
...
+## Introduction
+
+The TON Blockchain achieves consensus through a layered Byzantine Fault Tolerant (BFT) protocol designed for **high throughput, low latency, and security against up to one-third malicious validators**.
+
+The system combines two protocols:
+
+- **Catchain Protocol** – a reliable broadcast mechanism that ensures consistent message delivery and fork detection.  
+- **Block Consensus Protocol (BCP)** – a three-phase commit protocol built on top of Catchain that finalizes blocks.  
+
+This separation allows TON to finalize blocks in **3–6 seconds** with **hundreds of validators** across the globe.
```
reveloper: This was improved with TON node engine upgrades.
- Block preparation around 1 sec
- Masterchain block finality up to 3 sec

reveloper:
@laviniat1996 Please, take into account the following info

## Audience
This page is designed for individuals (developers) seeking to understand how consensus is maintained in TON and how security is guaranteed.  We can gain additional insights from the top-performing pages of other chains. I noticed that readers who seek this typically acknowledge this for other chains.

## Scope of work
- This document should cover the questions of how consensus works in TON.
- Catchain itself doesn't cover the entire consensus flow, but this is the fundamental base for building this page. 
- Validator elections are not related to the consensus mechanism and should be described in a separate page.

## Technical insights for rework
Here is the feedback and suggestions I collected from TON Core.

### 1. Explain what Catchain is?
The Catchain core thing is to build a DAG tree of options, catch deviation events (that’s why it was initially called **catch** **chain**), and adjust the difference.

```md
Catchain Overview

Catchain is a communication protocol between validators. It does not execute the consensus algorithm itself but prepares data required for the decision-making of a higher-level component: validator-session (BCP). Essential catchain tasks are:

- setting up a network overlay on the top of ADNL ensuring communication between validators;
- setting up and updating a list of neighbor nodes for direct communications inside the overlay;
- receiving catchain blocks from other validators and sending back decisions of the current validator;
- controlling blocks dependencies;
- managing an internal database holding current catchain session results
- restoring catchain session state after a validator restart;
- detecting forks and blaming validators;
- maintaining the consensus algorithm.
```

### 2. Make a clear definition and roles of Catchain and BCP in TON
Unlike Classic three-phase BFT(pre-prepare, prepare, commit) in current PR, TON uses Catchain DAG. Overall this works according the following steps:
  - Validators exchange _block candidates_
  -  DAG synchronization with causal ordering is used
  - Nodes sign only one branch, otherwise - blaming (marking a participant as a bad actor, the entire chain from them is invalidated, and they are banned)
  - Consensus is reached without all-to-all voting
  - The decision on what to do with blamed participants performed by BCP, it's not the scope of the catchain
  
###  3. Explain the definition and the difference between a catchain session and и validator session.

 ##  Important references
  - More detailed reference: [link](https://docs.ton.dev/86757ecb2/p/07ddda-walk-through-the-catchain)
  - Example implementation of how 11 different nodes generate a random number and then agree through catchain which number is better to choose (in this case, whoever has the bigger number is right): [link](https://github.com/ton-blockchain/ton/blob/master/test/test-catchain.cpp)

anton-trunov:
@Shvandre will take care of issue #171

PR #492, opened 2025-09-26
sharding

skywardboundd:


novusnota:
Oh no! After a second careful review, I think this should be rewritten because it does not facilitate learning and applying this pattern in practice.

PR #541, opened 2025-10-01
feat: config descriptions

aigerimu:
closes https://github.com/tact-lang/mintlify-ton-docs/issues/481


pyAndr3w (approved):
LGTM — no comments on the PR itself.

However, the article contains several `ConfigParams` that are not present in the current blockchain configuration, and some that are available only on the testnet. To avoid confusion, these parameters should either be clearly marked or removed from the documentation.

Kaladin13 (approved):

PR #604, opened 2025-10-07
FunC: functions page

### LINKED ISSUE #711: [Language > FunC] Functions
Fix the functions page

jeshecdom:
Closes #711.

Some links are still missing, as I do not know where that info is in the new docs. The missing links are described in issue #712.


novusnota:
> (maybe I will copy info from Tact docs regarding asm functions).

very nice

latest source: https://raw.githubusercontent.com/tact-lang/tact/refs/heads/main/docs/src/content/docs/book/assembly-functions.mdx

language/func/asm-functions.mdx
```diff
@@ -0,0 +1,86 @@
+---
+title: "Assembler functions"
+sidebarTitle: "Assembler functions"
+noindex: "true"
+---
+
+import { Aside } from '/snippets/aside.jsx';
+
+<Aside type="danger">
+    Needs extensive revamp. Maybe use Tact docs as guidance.
```
verytactical: Please don't put TODO in code

language/func/asm-functions.mdx
```diff
@@ -0,0 +1,86 @@
...
+
+## Assembler function body definition
+
+In FunC, functions can be defined directly using assembler code. This is done using the `asm` keyword, followed by one or more assembler commands written as strings.
+For example, the following function increments an integer and then negates it:
+
+```func
+int inc_then_negate(int x) asm "INC" "NEGATE";
+```
+– a function that increments an integer and then negates it. Calls to this function will be translated to 2 assembler commands `INC` and `NEGATE`. An alternative way to define the function is:
```
verytactical: Duplicate to

> the following function increments an integer and then negates it:

verytactical: > An alternative way to define the function is:

There is no code after this sentence.

language/func/functions.mdx
```diff
@@ -1,257 +1,225 @@
...
-A function name can be any valid [identifier](/language/func/literals_identifiers#identifiers). Additionally, it may start with the symbols `.` or `~`, which have specific meanings explained in the [Statements](/language/func/statements#methods-calls) section.
+where `[ ... ]` represents an optional entry. Here, 
 
-For example, `udict_add_builder?`, `dict_set`, and `~dict_set` are all valid function names, and each is distinct. These functions are defined in [stdlib.fc](/language/func/stdlib).
+- `<forall declarator>` specifies that it is a polymorphic function. This is optional.
+- `<return_type>` is the return type of the function.
+- `<function_name>` is the function identifier.
+- `<comma_separated_function_args>` is a comma separated list of function arguments, each argument declaring a type and an identifier.
+- `<specifiers>` are specifiers that instruct the compiler on how to process the function.
+- `<function_body>` is the actual function body, which can be of three kinds: a function declaration (i.e., no body), an assembler function, or a standard function body definition.
```
verytactical: "For example:" and several examples of such a declaration

languages/func/functions.mdx
```diff
@@ -1,257 +1,225 @@
...
+- `pair_swap([2, 3])` returns `[3, 2]`;
+- `pair_swap([1, [2, 3, 4]])` returns `[[2, 3, 4], 1]`.
 
-### Return type
+In this example, `X` and `Y` are [type variables](/language/func/types#polymorphism-with-type-variables). When the function is called, these variables are replaced with actual types, and the function executes accordingly. Even though the function is polymorphic, the compiled assembly code remains the same for any type substitution. This is possible due to the polymorphic nature of stack manipulation operations. However, other forms of polymorphism, such as `ad-hoc` polymorphism with type classes, are not currently supported.
+
+It is important to note that `X` and `Y` must each have a type width of 1, meaning they should fit within a single stack entry. This means you can't use `pair_swap` on a tuple like `[(int, int), int]` because type `(int, int)` has a width of 2, taking up two stack entries instead of one.
+
+
+## Return type
```
verytactical, line 25: Return type is probably used a bit more than forall :)

Please put it above that section

language/func/functions.mdx
```diff
@@ -1,257 +1,225 @@
...
 
-FunC also supports **type inference**. For example:
+FunC also supports **type inference** with the use of underscore `_` as the return type. For example:
 ```func
 _ pyth(int m, int n) {
   return (m * m - n * n, 2 * m * n, m * m + n * n);
 }
 ```
-This is a valid definition of the function `pyth`, which has the inferred type `(int, int) → (int, int, int)`.
+This is a valid definition of the function `pyth`, which has the inferred type `(int, int) -> (int, int, int)`.
```
verytactical: It is, but pythagorean triples are probably not very much used in smart contracts.

An example of gas calculation or divmod would be more fitting.

language/func/functions.mdx
```diff
@@ -1,257 +1,225 @@
...
-example(a);
-a.example();
-```
+Every program must include a function with `id = 0`, meaning it must define either `recv_internal` or `main`, but **not** both.
+The `run_ticktock` function is used in ticktock transactions of special smart contracts.
 
-A function with at least **one argument**, it can be called a **non-modifying method**. For example, the function `store_uint` has the type `(builder, int, int) → builder`, where:
-- The second argument is the value to store.
-- The third argument is the bit length.
+#### Receive internal
```
verytactical: I think it's not a section about a special function name per se, but about that special function in general.

Probably this should go into a separate section/article about entry points.

languages/func/functions.mdx
```diff
@@ -301,11 +269,11 @@ Another example of using `inline` from [ICO-Minter.fc](https://github.com/ton-bl
 }
 ```
 
-#### `inline_ref` specifier
+### `inline_ref` specifier
```
verytactical, line 180: This section could use a little bit more clarity on how exactly this is implemented. Probably, add an example with the code that uses `inline_ref` function twice, and how exactly it's stored in BoC.

language/func/functions.mdx
```diff
@@ -301,11 +269,11 @@ Another example of using `inline` from [ICO-Minter.fc](https://github.com/ton-bl
 }
 ```
 
-#### `inline_ref` specifier
+### `inline_ref` specifier
 
 When a function is marked with the `inline_ref` specifier, its code is stored in a separate cell. Each time the function is called, TVM executes a `CALLREF` command. This works similarly to `inline`, but with a key difference—since the same cell can be reused multiple times without duplication, `inline_ref` is generally more efficient regarding code size. The only case where `inline` might be preferable is if the function is called just once. However, recursive calls to `inline_ref` functions remain impossible, as TVM cells do not support cyclic references.
 
-#### `method_id`
+### `method_id`
```
verytactical: The explanation probably would take a whole section of documentation: we have to explain what a method dictionary is, and how incoming messages use message id to find their handler.

verytactical: > avoid setting method IDs manually

Should also mention a very common case when it's needed: when method id is set by TEP and doesn't match a generated id.

language/func/functions.mdx
```diff
@@ -366,106 +334,32 @@ int get_counter() method_id {
...
-- `pair_swap([1, [2, 3, 4]])` returns `[[2, 3, 4], 1]`.
-
-In this example, `X` and `Y` are [type variables](/language/func/types#polymorphism-with-type-variables). When the function is called, these variables are replaced with actual types, and the function executes accordingly. Even though the function is polymorphic, the compiled assembly code remains the same for any type substitution. This is possible due to the polymorphic nature of stack manipulation operations. However, other forms of polymorphism, such as `ad-hoc` polymorphism with type classes, are not currently supported.
-
-It is important to note that `X` and `Y` must each have a type width of 1, meaning they should fit within a single stack entry. This means you can't use `pair_swap` on a tuple like `[(int, int), int]` because type `(int, int)` has a width of 2, taking up two stack entries instead of one.
-
+## Function body 
 
-## Assembler function body definition
+### Function declaration
```
verytactical: Should probably mention why do we need them.

languages/func/functions.mdx
jeshecdom: I think that at the end I decided to move forall to the end of the article.

language/func/functions.mdx
jeshecdom: Agreed, I had that thought when I read the example.... but left it there. I'll change it.

language/func/functions.mdx
jeshecdom: You mean that I should just list the names of the special functions, and then point to a separate article explaining the special functions handling TVM events?

languages/func/functions.mdx
jeshecdom: I will need to ask about this one. I actually do not understand fully what it is doing.

language/func/functions.mdx
jeshecdom: Yes, agreed.

language/func/functions.mdx
jeshecdom: Again, I will ask about this.

languages/func/functions.mdx
verytactical: Good idea :)

language/func/functions.mdx
verytactical: Yes, exactly that

languages/func/functions.mdx
verytactical: So from bitcode point of view there are several options to refer to a function (i.e. continuation, represented as some DAG of cells):
- Link it into a function dictionary. Now function calls look up the dictionary by function id and run whatever they found there.
- Link the root cell of function directly to the instruction that calls it. Due to cell deduplication during BoC serialization, all the call sites referring to the same root cell will actually link to the same node in a cell DAG. I think this is `inline_ref`.
- Code of function might be very small, only a single cell, but we'll still have to open the cell and pay for it. So instead of referring to the root cell of a function, we glue bits of that root cell directly into the cell that calls it, and use another asm instruction that runs the continuation from bits (instead of ref). This is likely `inline`.

jeshecdom: Please have a look at the explanation I added for `inline_ref`.

language/func/functions.mdx
jeshecdom: I did not include the example for TEPs, because they specify the opcodes of messages, not the method_ids of functions in the implementation.

novusnota:
Re-deployed: https://companyname-a7d5b98e-jeshecdom-functions.mintlify.app/

PR #616, opened 2025-10-08
feat: links to original pdfs from web versions of whitepapers

### LINKED ISSUE #567: [Whitepapers] Add links to the original pdfs
It should be done at the top of each whitepaper with a warning the web versions contain some fixes (with a warning near each fix).

novusnota:
Closes #567

Here's an example:

image

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -300,6 +300,7 @@
                   "standard/tokens/jettons/overview",
                   "standard/tokens/jettons/how-it-works",
                   "standard/tokens/jettons/comparison",
+                  "standard/tokens/jettons/how-to-transfer",
                   "standard/tokens/jettons/how-to-get-supply-data",
                   "standard/tokens/jettons/mintless/overview",
                   "standard/tokens/jettons/mintless/how-to-deploy"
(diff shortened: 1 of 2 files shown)


novusnota:
There are spelling issues in the whitepapers, attempting to fix all...

UPD: I was also re-doing all the regex ignores as part of the #617, but that task helped discover that almost none are needed, actually. However, there were edge-cases I could not resolve easily enough, and fixing spelling in all files **should be one of the last tasks** before the actual public release.

novusnota:
### TL;DR: This PR is complete despite false-positives in spellchecks.

There are false-positives in whitepaper spellchecks and they cannot be ignored with `{/* vale off */} ... {/* vale on */}` comments due to some weird issues with Vale when parsing larger files. Like, JSX expressions `{JS code}` that are placed very early in the file cause some weird regexes (or Vale's parse engine) to fail about 200 lines deeper.

I've tried to debug the issue and modify the configuration (#617) to significantly reduce the number of RegExes, aiming to decrease the rate of false positives and uncover possible overlooked spelling errors. There were a couple of issues with that: small reproducible samples and low confidence in RegExes (e.g., they might not encompass all cases, but they should not induce false-positives).

I'll resume work on that later as per #617 as this is not a pressing issue for the time being.

PR #687, opened 2025-10-13
FunC global variables and compiler directives pages

### LINKED ISSUE #685: [Language > FunC] Compiler Directives
Fix Compiler directives page

### LINKED ISSUE #686: [Language > FunC] Global Variables
Fix global variables page

jeshecdom:
Closes #686. Closes #685.

### DIFF
--- a/.github/scripts/build_review_instructions.py
+++ b/.github/scripts/build_review_instructions.py
@@ -28,13 +28,27 @@ Scope and priorities:
 2. Only after style compliance, check for obvious, provable, blocking errors not covered by the guide (e.g., an incorrect calculation or an unsafe, non‑runnable step) and report them with proof. If not certain from repo content alone, omit.
 
 Review protocol:
-- Inspect only .md/.mdx files touched by this PR. Ignore all other file types.
+- Inspect only content files touched by this PR: `.md`, `.mdx`, and `docs.json`.
+- It is acceptable to report findings that originate in `docs.json` (e.g., broken or duplicate paths/slugs, invalid sidebar grouping, typos in titles). When the problem is in `docs.json`, cite its exact lines.
 - Examine only the lines changed in this diff (use surrounding context as needed). Do not flag issues that exist solely in unchanged content.
 - Report every issue you see in this diff; do not postpone or soften problems.
 - Location links must be repo-relative paths such as pending/discover/web3-basics/glossary.mdx?plain=1#L10-L12 (no https:// prefix).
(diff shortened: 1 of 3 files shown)


languages/func/compiler-directives.mdx
```diff
@@ -15,7 +15,7 @@ The `#include` directive enables the inclusion of another FunC source file parse
 **Syntax:**
 
 ```func
-#include "filename.fc";
+#include "<path_to_filename>";
```
verytactical, line 20: Does it allow cyclic includes, i.e. when some file imports itself, directly or indirectly?

languages/func/global-variables.mdx
```diff
@@ -1,79 +1,107 @@
 ---
-title: "FunC global variables"
+title: "Global variables"
 sidebarTitle: "Global variables"
 noindex: "true"
 ---
 
-A FunC program primarily consists of function declarations/definitions and global variable declarations. 
-This section focuses on the latter.
+## Declaration
```
verytactical, line 20: This is a nice reference article, except it doesn't answer several key questions:
- What does it mean for a variable to be global? (i.e. it's stored in a certain control register as part of tuple)
- Why would someone want to make a variable global?
- What if there are 2 global variables with the same in different files?
- What if the same file with a global variable is included twice?

languages/func/global-variables.mdx
```diff
@@ -1,79 +1,107 @@
...
 ```
 
-In this example, the global variable `op` is assigned the addition operator `_+_`. The program then verifies the associativity of addition using three sample integers: 2, 3, and 9.
+The example defines two auxiliary functions: `check_commutative` and `add`, in addition to the program entry point [`main`](#special-functions#main).
+Function `check_commutative` checks if the operator stored in the global variable `op` satisfies the commutative property for the specified inputs `a` and `b`.
+Function `add`, adds its two inputs.
+Function `main` assigns the addition function `add` to the global variable `op`. Then, it verifies the commutativity of addition for the specific values: `2`, `3`.
 
-Under the hood, global variables in FunC are stored in the `c7` control register of the TVM, with a maximum limit of 31 variables.
+In FunC, you can _omit the type_ of global variables. In this case, the compiler determines the type based on how the variable is used. Equivalently,
```
verytactical, line 54: _nit:_ I think this is applicable to arbitary variable declarations, and should just link to a corresponding section of documentation about variables.

jeshecdom:
Included suggested fixes and some of those suggested by AI. Pending issues described in:

#863
#864

languages/func/global-variables.mdx
jeshecdom: I did not do this because local variables and global variables are completely different entities in FunC. Local variable declarations are expressions, while global variable declarations are not. There are also other differences, for example global variables can omit the type, but local variables do not allow this. So, I prefer to leave them on their own page independently one from another.

verytactical (approved):

PR #725, opened 2025-10-14
chore: increase maximum width of the content area from 42 to 48rem

novusnota:
This is a rather safe change because:

1. We're changing `max-width`
2. It's changed by 6rem units, which aligns with other width constraints Maple theme has
3. We're not pushing the limits of `max-width` and not overlapping navigation nor table of contents (yet)


novusnota:
After some internal discussion, this change shall be put on hold

PR #738, opened 2025-10-14
feat: stub component

### LINKED ISSUE #739: Put message on stub pages
Stub pages merely have a word "Stub".

Create a component that shows a human-readable message, and an optional link to an issue that tracks progress on a page

verytactical:
Closes #739

### DIFF
--- a/guidebook/debug.mdx
+++ b/guidebook/debug.mdx
@@ -1,5 +1,5 @@
-## title: "Hands-on debugging"
-
+---
+title: "Hands-on debugging"
 ---
 
 import {Aside} from "/snippets/aside.jsx";


snippets/stub.jsx
```diff
@@ -0,0 +1,147 @@
+export const Stub = ({ issue }) => {
+    // Obviously neither imports not variables in Mintlify snippets do not work :)
```
novusnota, line 2: @verytactical Well, yes, but we can actually use other snippets or components without an explicit import inside `.jsx`. We just need to import all on the resulting page.

For example, here's the MDX:

```mdx
---
title: "Start here"
---

import { Stub } from '/snippets/stub.jsx';

<Stub issue="42" />
```

And here's the `/snippets/stub.jsx`:

```jsx
export const Stub = ({ issue }) => {
  const issueEl = issue ? (
    <>
      You can track progress on this page in{" "}
      <a href={`https://github.com/tact-lang/mintlify-ton-docs/issues/${issue}`}>issue #{issue}</a>.
    </>
  ) : null;

  return (
    // @ts-ignore
    <Aside>
      This page is a placeholder. {issueEl}
      {/* @ts-ignore */}
    </Aside>
  );
};
```

This renders nicely:

image

verytactical: We won't follow along Mintlify bugs.

novusnota: Considering that Aside will be present on most pages anyways, this shouldn't be that much of a hack :)

novusnota: > We won't follow along Mintlify bugs.

Hmm, but that's not a bug — otherwise the components such as `<Icon>` that are "built-in" to Mintlify wouldn't be working here.

Or did you mean that this behavior is Mintlify-specific?

PR #875, opened 2025-10-15
feat: add ecosystem RPC overview

### LINKED ISSUE #159: [Ecosystem > API > Overview]
Old article:
https://docs.ton.org/v3/guidelines/dapps/apis-sdks/api-types

- comparison table of popular services
  - rows (services)
    - public liteservers
    - toncenter v2
    - toncenter v3
    - tonapi
    - dton
    - self-hosted liteserver
    - self-hosted toncenter v2
    - self-hosted toncenter v3
  - columns (properties)
    - is self-hosted
    - is open-source
    - is indexing?
    - is archival?
    - has proofs?
    - (testnet/mainnet endpoints) / (source code + deploy guide)
    - documentation

laviniat1996:
Closes #159

This PR adds an overview page for the ecosystem RPC section:
- explains the three types of APIs available for fetching data on TON
- the main diffs between them (http vs indexer, adnl vs http)
- tells devs what the best use is for each one
- documents official and third party providers, including docs links and enpoints

### DIFF
--- a/standard/tokens/metadata.mdx
+++ b/standard/tokens/metadata.mdx
@@ -189,11 +189,11 @@ Read known attributes such as `uri`, `name`, `image`, `description`, and `image_
 
 Examples:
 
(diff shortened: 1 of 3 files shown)


reveloper:
@laviniat1996 
I recommend arranging the structure closer to the original [APIs](https://docs.ton.org/v3/guidelines/dapps/apis-sdks/api-types/) overview, as readers should first find possible ways to build their solutions and then access more explanations on types differences.

All APIs listed here are "official" within the TON Ecosystem, so we shouldn't categorize them as official / non-official.
Instead, we can group them by type and then provide a more detailed explanation of what each type means.

This API list was reviewed just a month ago. The problem point on I suggested SDKs, I mentioned @ton/ton, because we have no other choice at the moment.

cc @verytactical

PR #882, opened 2025-10-15
fix(landing): decrease font size for smaller displays

novusnota:
A follow-up to #870 that corrects the font sizes for smaller displays, especially for mobile previews. That's important because the announcement post would most likely be opened from mobile devices in most cases.

### DIFF
--- a/resources/dictionaries/custom.txt
+++ b/resources/dictionaries/custom.txt
@@ -232,6 +232,7 @@ SDKs
 SECD
 SELinux
 STON.fi
+SVM
 Scannability
 Schnorr
 Sedov


index.mdx
```diff
@@ -51,7 +51,7 @@ export const JourneyStep = ({ title, icon, iconType, href, cta, children }) => {
 };
 
 <div
-  className="relative w-full box-border flex flex-col mx-auto px-6 max-w-xl md:max-w-3xl lg:max-w-4xl xl:max-w-5xl"
+  className="relative w-full box-border flex flex-col mx-auto px-6 max-w-xl sm:max-w-2xl md:max-w-3xl lg:max-w-4xl xl:max-w-5xl"
```
verytactical, line 54: Let's remove all the HTML markup from here. We are not authors of Mintlify, and all of this can silently break at any release on their side, any time of the day.

verytactical (changes requested):
.

index.mdx
novusnota: This page has `mode: "custom"` above, none of the theme styles are applied apart from colors (only when specified explicitly) and inner styles of existing components (as expected)

novusnota: I'm literally changing my styles that I've applied on top of none :)

The only concern here might be the switch from Tailwind v3 (current) to Tailwind v4 (or any other framework), but in that case everything will break everywhere, including many other customers who use Tailwind styling in production like us. Therefore, I don't see that as a big concern.

If there ever comes a day when Mintlify breaks all its customers, we'll just start to look into Astro more eagerly

verytactical: We shouldn't have applied any in the first place without figuring out process issues with Mintlify.

Issue #663, opened 2025-10-10
[Wallets > Highload wallets] Provide images for "Message sending flow"


verytactical:
Highload wallet v3 specification article has "Message sending flow" graph in ASCII art. Replace it with an image.

PR #887, opened 2025-10-15
feat(wallet contracts): better illustrations

aigerimu:
closes https://github.com/ton-org/docs/issues/663


anton-trunov:
@aigerimu how is the progress here?

PR #905, opened 2025-10-17
chore: add GitHub repo link to the top right area of each article

### LINKED ISSUE #884: Add Github icon on the main page or in title nav
There must be a link to GH everywhere. Documentation users expect it to be there.

novusnota:
That is, except for the `index.mdx`. Also notice that "Edit suggestions" and "Raise issues" buttons would soon be enabled by the end of each page (except `index.mdx`) via the Mintlify's dashboard panel.

Closes #884.

image

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -24,11 +24,17 @@
   },
   "contextual": {
     "options": [
+      {
+        "title": "GitHub",
+        "description": "Open GitHub repository",
+        "href": "https://github.com/ton-org/docs",
+        "icon": "github"
+      },
       "copy",
+      "mcp",
       "chatgpt",
       "claude",
       "perplexity",
-      "mcp",
       "cursor",
       "vscode"
     ]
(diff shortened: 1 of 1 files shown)


docs.json
```diff
@@ -24,11 +24,17 @@
   },
   "contextual": {
     "options": [
+      {
+        "title": "GitHub",
```
verytactical, line 28: All the other items in the list start with a verb.

The link to the root of the repo is misleading here, because it's the button for the actions to be done with this article precisely.

I've started a conversation with Mintlify team to figure out a way to handle `$path` here.

verytactical (changes requested):

novusnota:
There are buttons for suggesting edits or raising issues towards any given page. They are placed at the bottom, right after the page content ends.

image

These buttons solve the issue this PR was made to address, albeit in a different way. Thus, this PR will be closed. Yet, the corresponding issue will be marked as "not planned" — we may or may not come back to it in the future, but the goal has been achieved.

PR #906, opened 2025-10-17
feat: require stack and set default empty array for runGetMethod

### LINKED ISSUE #467: [Ecosystem > API > TON Center > API v2 > Smart contracts > Run get-method] Fix interface
In our playground, we can’t send a stack without any variables ("stack": []), but it is required, for example, in get_nft_data when no arguments are provided.

If the user doesn’t click on “Add an item”, we should automatically add "stack": [].

laviniat1996:
closes #467 

Updated V2RunGetMethodRequest schema:
  - Added `stack` to required fields.
  - Set `default: []` for `stack` to ensure the Playground includes it automatically.


novusnota (dismissed):

verytactical (changes requested):
Now there is no description whatsoever of what is the format of that field. Please fix `V2StackEntity` type along with this change.

ecosystem/rpc/toncenter/v3.yaml
```diff
@@ -1477,40 +1477,7 @@ paths:
...
-        "type": "slice",
-
-        "value": "te6..." // base64 encoded boc with slice
-
-        }
-
-        ]
-
-        ```'
+      description: 'Executes a read only get-method of a smart contract. Use this endpoint to call TVM methods without creating a transaction onchain. The request must include the method name and "stack" array with input arguments.'
```
anton-trunov, line 1480: there is no such thing as "read only get-method"

anton-trunov (changes requested):

ecosystem/rpc/toncenter/v3.yaml
novusnota: `s/ read only//`

novusnota:
Thanks for the work! That said, changes made here diverge from the upstream TON Center API v3 specification. A proper solution requires careful changes to the upstream code, followed by a port of the resulting specification back to the docs.

https://github.com/toncenter/ton-indexer/blob/bd15a121b3a450f34546176ada3ca475db045704/ton-index-go/main.go#L1805-L1834

I'll close this PR, but will preserve the branch so some parts can be reused later. Thanks again!

PR #916, opened 2025-10-18
fix: jetton transfer code

### LINKED ISSUE #915: [Jetton -> How to: Transfer]
Quickly describe what is response_destination and set it to sender address in example

And fix provider(destAddress) it should be from wallet

Image

Karkarmath:
Closes #915


verytactical (changes requested):
We did a review with @anton-trunov and pushed most of the changes we'd like to see into the branch. There are several more things we didn't have time to do, but are required to move forward with this article.

- Make it absolutely explicit in the name of variables which wallets are regular and which are Jetton wallets;
- Low-level example takes 5 addresses, but high-level takes only 2. They have to be consistent to each other. Make the low-level example use precisely the same parameters.
- Check that examples are executable. Record a screencast, then convert the video into an explanation how it can be done. Put this explanation into the article.

PR #923, opened 2025-10-20
feat(api): generate MDX files for API pages

### LINKED ISSUE #660: Consider auto-generating MDX files for API pages
_created due to some compromises made in https://github.com/tact-lang/mintlify-ton-docs/pull/655_

This could help with:

1. Customization of said pages
2. Ordering
3. Linkage — otherwise, `npm run check:links` will flag links to API pages as broken

A rather big negative point of such change would be increase in maintenance efforts, because we'd have to keep OpenAPI JSON/YAML files and MDX files in sync, whereas with the current approach we only need to keep track of OpenAPI files. Though only the `openapi: get /...` frontmatter property is needed of such files, and this might not be so big of a deal, but still.

Things to keep in mind:
- Put the auto-generated MDX-for-API files in various `.ignore` files, except for `.gitignore`

Mintlify docs:
- https://www.mintlify.com/docs/api-playground/openapi-setup#create-mdx-files-for-api-pages

novusnota:
Closes #660

This change:

1. Alleviates issues with "broken links" directed at API routes
2. Allows adding how-tos and auxiliary documentation prefacing generated parts in API routes

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,8 @@
     "spell": "npm run check:spell",
     "spell:some": "cspell --show-suggestions",
     "lint:all": "npm run check:all",
+    "gen:mdx-for-openapi:toncenter-v2": "npx @mintlify/scraping@latest openapi-file -w ecosystem/rpc/toncenter/v2.json -o ecosystem/rpc/toncenter/v2",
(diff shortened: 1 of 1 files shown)


verytactical (approved):
Leftovers separated to #950

Issue #976, opened 2025-10-27
[Language > FunC > Types] Improve article


aigerimu:
https://gist.github.com/aigerimu/4460358d66fc0d8bc4a7fced1aba2f14

Issue #299, opened 2025-09-18
[Language > FunC] Types


jeshecdom:
1. In this paragraph:

    > In FunC, the `null` value of the TVM type `Null` represents the absence of a value for a given atomic type. While `null` is generally not allowed, some standard library functions handle it in specific ways:
    
    "`null` is generally not allowed"... so, When is `null` allowed outside of the specific cases mentioned for library functions? Is it possible for a programmer to handle the `null` value? How is it done? An example is needed here. 

2. In this paragraph:
    
    > Additionally, an atomic type A can be implicitly transformed into A^? (also known as Maybe A), allowing a variable of type A to store either a valid value or null. This transformation happens automatically and is not enforced by the type checker.
    
    An example is needed that shows this implicit transformation. Also, it is not clear if a programmer can declare a variable to be of type `A^?`.

3. Section on "Hole type". It is given an example of usage for `var`, but no example for use of `_`.

4. Section on "Functional type". 

    > Internally, values of functional types are represented as continuations.
    
    It is better not to mention low-level technicalities if no example is provided that illustrates the new concept, i.e., what is a continuation? and how functions are represented as continuations? In fact, deleting the "continuations claim" makes the section clearer. If it is a must that the continuation claim should be in the doc, it could be placed in a separate document that explains low-level details about the language.
    
5. Another observation that is all over TON docs: Do not replace ASCII characters with special symbols. For example, do not replace `A -> B` with `A → B`. I've seen these replacements happening in code snippets!

6. Section on "Tensor types". 
    
    The following statement should not be under "Examples", but after the description of the unit type:
    
    > The unit type () has a single value, also written as (), occupying zero stack entries.

    The following two sentences are saying the same thing. Remove the first one, because it does not sound right that it should occur under "Examples":

    > Type of form (A) is considered by type checker as the same type as A.   

    > Note: A type written as (A) is treated as identical to A by the FunC type checker.
    
7. Section "Polymorphism with type variables".

    Remove this paragraph, because the next paragraph is saying the SAME thing:
    
    > is a polymorphic function which takes a (single stack entry) value and returns two copies of this value. duplicate(6) will produce values 6 6, and duplicate([]) will produce two copies [] [] of an empty tuple.
    
    The syntax for polymorphic functions MUST be explained. It should be said that the polymorphic variables occur to the left of `->`. And to the right of `->` is the usual function declaration (which by the way, it was was NOT explained in the "Functional type" section... :P)
    
8. Section "Type width".

[...]

PR #978, opened 2025-10-27
feat: [Language > FunC > Types] Improve article

aigerimu:
closes https://github.com/ton-org/docs/issues/976 and #299

### DIFF
--- a/languages/func/types.mdx
+++ b/languages/func/types.mdx
@@ -12,7 +12,7 @@ FunC includes built-in types forming the language’s core.
 
 - `int` is a 257-bit signed integer type. Overflow checks are enabled by default and trigger an exception if exceeded.
 
-- `cell` is a [TVM cell type](/tvm/serialization/cells) used to store persistent data in TON Blockchain. Data is organized in a [bag of cells](/tvm/serialization/boc), with each cell containing up to 1023 bits of arbitrary data and up to four references to other cells.
+- `cell` is a [TVM cell type](/foundations/serialization/cells) used to store persistent data in TON Blockchain. Data is organized in a [bag of cells](/foundations/serialization/boc), with each cell containing up to 1023 bits of arbitrary data and up to four references to other cells.
 
 - `slice` is a read-only view of a cell that allows sequential access to its data and references. A cell can be converted into a slice, extracting stored bits and references without modifying the original cell.
 


languages/func/types.mdx
```diff
@@ -6,65 +6,70 @@ noindex: "true"
...
 
-FunC includes several built-in types that serve as the foundation of the language.
+FunC includes built-in types forming the language’s core.
 
 ## Atomic types
 
 - `int` is a 257-bit signed integer type. Overflow checks are enabled by default and trigger an exception if exceeded.
 
-- `cell` is a TVM cell type used to store persistent data in the TON Blockchain. Data is organized in trees of cells, with each cell containing up to 1023 bits of arbitrary data and up to four references to other cells. Cells function as memory units in stack-based TVMs.
+- `cell` is a [TVM cell type](/foundations/whitepapers/tvm#1-1-3-preliminary-list-of-value-types) used to store persistent data in TON Blockchain. Data is organized in a [bag of cells](/tvm/serialization/boc), with each cell containing up to 1023 bits of arbitrary data and up to four references to other cells.
```
anton-trunov: in general, we should not link to the whitepapers, prefer the corresponding docs page

languages/func/types.mdx
```diff
@@ -6,65 +6,70 @@ noindex: "true"
...
 
 - `slice` is a read-only view of a cell that allows sequential access to its data and references. A cell can be converted into a slice, extracting stored bits and references without modifying the original cell.
 
 - `builder` is a mutable structure used to construct cells by adding data and references before finalizing them into a new cell.
 
 - `tuple` is an ordered collection of up to 255 elements, each capable of holding a value of any type.
 
-- `cont` is a TVM continuation used to manage execution flow in TVM programs. Although a low-level construct, it provides flexible execution control.
+- `cont` is a TVM continuation used to manage execution flow in [TVM instructions](/tvm/instructions).
```
anton-trunov: actually, there should be a dedicated TVM continuation page

cc @hacker-volodya

languages/func/types.mdx
```diff
@@ -6,65 +6,70 @@ noindex: "true"
...
 
-Each of these types occupies a single slot in the TVM stack.
+- `Null`  is a type with a single value, used to represent empty lists or the absence of a return value.
```
anton-trunov: `Null` does not represent the absense of a return value: if a function does not return a value it just does not push anything on the stack

languages/func/types.mdx
```diff
@@ -6,65 +6,70 @@ noindex: "true"
...
-FunC does not have a dedicated boolean type.
+FunC does not have a boolean type.
 Instead, booleans are represented as integers:
-- `false` is `0`, `true` is `-1` (a 257-bit integer with all bits set to 1).
-- Logical operations are performed using bitwise operations.
-- In conditional checks, any nonzero integer is treated as `true`.
+
+- `false` is `0`, `true` is `-1`, a 257-bit integer with all bits set to 1.
+- Logical operations are performed using [bitwise operations](/languages/func/operators).
+- In [conditional checks](/languages/func/statements#conditional-statements), any nonzero integer is treated as `true`.
```
anton-trunov: in addition to the conditional statement, these conditional checks are present in the loop conditions and the conditional expression

languages/func/types.mdx
```diff
@@ -73,103 +78,116 @@ Types can be combined to form more complex structures.
...
 
-### Tensor types
+<Aside>
+  Unlike the unit type `()`, an empty tuple `[]` occupies one stack entry.
+</Aside>
+
+### Tensor type
 
-Tensor types represent ordered collections of values and are written in the form `(A, B, ...)`.
+Tensor types represent collections of values with a fixed order and are written in the form `(A, B, ...)`.
```
anton-trunov: the original version is better ("ordered collections" is a common term in computer science and maths)

languages/func/types.mdx
```diff
@@ -73,103 +78,116 @@ Types can be combined to form more complex structures.
...
 These types occupy multiple TVM stack entries, unlike atomic types, which use a single entry.
 
 **Example:**
-If a function `foo` has the type `int -> (int, int)`,
-it takes one integer as input and returns a pair of integers as output.
-A call to this function may look like: `(int a, int b) = foo(42);`.
+
+A function `foo` of type `int -> (int, int)` takes one integer as input and returns a pair of integers as output.
```
anton-trunov: @jeshecdom we need to clarify what "pair" means in this context: it's ambiguous

languages/func/types.mdx
```diff
@@ -73,103 +78,116 @@ Types can be combined to form more complex structures.
...
+(int a, int b) = foo(42);
+```
+
 Internally, the function consumes one stack entry and produces two.
 
-**Type representation:** Although the values `(2, (3, 9))` of type `(int, (int, int))` and `(2, 3, 9)` of type `(int, int, int)` are stored identically as three stack entries `(2, 3, and 9)`, FunC treats them as distinct types.
-For instance, the following code **will not compile**:
+**Type representation:**
+
+Values `(2, (3, 9))` of type `(int, (int, int))` and `(2, 3, 9)` of type `(int, int, int)` are stored identically as three stack entries `(2, 3, and 9)`.
```
anton-trunov: this does not make any sense: `(2, 3, and 9)`

languages/func/types.mdx
```diff
@@ -73,103 +78,116 @@ Types can be combined to form more complex structures.
...
 - A function takes no arguments
 
 The unit type `()` has a single value, also written as `()`, occupying **zero stack** entries.
 
 **Examples**
+
 - `print_int` has the type `int -> ()`, meaning it takes an integer but returns nothing.
 - `random` has the type `() -> int`, meaning it takes no arguments but returns an integer.
 
-### Tuple types
```
anton-trunov, line 124: why is this section removed?

languages/func/types.mdx
```diff
@@ -73,103 +78,116 @@ Types can be combined to form more complex structures.
...
+  The only exception is a tensor of a single element `(a)`, where `a` is not a tensor type itself. The compiler treats `(a)`as equivalent to `a`.
 </Aside>
 
 For more details, see the [Polymorphism with forall](/languages/func/functions#polymorphism-with-forall) section.
 
 ## User-defined types
 
-Currently, FunC does not support defining custom types beyond the type constructions described above.
-
-## Type width
```
anton-trunov, line 168: why is this section removed?

languages/func/types.mdx
jeshecdom: I just realized that type `Null` cannot be used in declarations in FunC, so I think that the best thing to do is to remove the "Null values" section and place the contents at the end of the "atomic types" section, by saying that all atomic types allow a special `null` value. And remove any mention of type `Null`, because it is too confusing. I can do this change.

languages/func/types.mdx
jeshecdom: Ok, I will add loops and conditional expressions.

languages/func/types.mdx
jeshecdom: Would it be ok to say?
> A function `foo` of type `int -> (int, int)` takes one integer as input and returns two integers as output, each one occupying a slot in the stack.

languages/func/types.mdx
jeshecdom: Maybe say?
> as three stack entries `2`, `3`, and `9`

languages/func/types.mdx
jeshecdom: It was moved before the tensor type. Conceptually, it is easier to explain first the tuple types, because they occupy a single slot in the stack and behave more like mathematical tuples. Then, later, once the tensor type is introduced, we can say that they represent ordered collections as with tuples, but the difference is that each element occupies a different slot in the stack. Although, there is some ambiguity with that statement, because `((2, 3), 5)`, the first element `(2, 3)` is actually two elements in the stack again.

languages/func/types.mdx
jeshecdom: This section was removed because I find out that after modifying the rest of the FunC docs, this concept of `type width` is actually never necessary to explain anything. Whenever it is needed, I found a way that was clearer to explain without invoking this concept. The concept is very ambiguous and difficult to define because it is not clear what it means by "this number is consistent for all values of a given type". It only confuses people.... hell, I am still confused by it.

languages/func/types.mdx
```diff
@@ -73,103 +78,116 @@ Types can be combined to form more complex structures.
...
+
+For example, `[int, cell]` defines a tuple with exactly two elements:
+
+- The first element is an integer.
+- The second element is a cell.
 
-Like in functional programming, it is possible to declare functional types which have in their domain and codomain other functional types. 
-For example, `(int -> int) -> int` is a function with domain `int -> int` and codomain `int`. 
-Similarly, `int -> (int -> int)` is a function with domain `int` and codomain `int -> int`.
+The type `[]` represents an empty tuple with a unique value—the empty tuple itself.
```
jeshecdom: > The type `[]` represents an empty tuple with a unique value—the empty tuple itself, which is also written as `[]`.

languages/func/types.mdx
```diff
@@ -73,103 +78,116 @@ Types can be combined to form more complex structures.
...
 
-### Tensor types
+<Aside>
+  Unlike the unit type `()`, an empty tuple `[]` occupies one stack entry.
```
jeshecdom: Mmmm... there is a reference to the unit type `()` here, but tensor types have not been explained yet.

languages/func/types.mdx
anton-trunov: Sometimes the FunC compiler mentiones "type width", afair. It's definitely worth it to explain it somewhere

languages/func/types.mdx
anton-trunov: sure

languages/func/types.mdx
anton-trunov: we have it now: https://beta-docs.ton.org/tvm/continuations

languages/func/types.mdx
jeshecdom: Ok, I put back the type width section.

PR #1026, opened 2025-10-29
feat(styles): sticky table headers

### LINKED ISSUE #1013: [Ecosystem > SDKs] Sticky table header
Image

When you scroll down, the SDKs table header is not visible anymore which makes it hard to read the table's content.

novusnota:
Closes #1013

### DIFF
--- a/extra.css
+++ b/extra.css
@@ -43,3 +43,19 @@ td>a>code {
   /* Firefox */
   scrollbar-width: none;
 }
+
+/* Sticky table headers */
+
+table {
+  overflow: unset !important;
+}
+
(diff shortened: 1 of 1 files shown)


novusnota:
> Would be nice to have some borders for table headers

Yeah, but that goes beyond what we can do without introducing new React components for tables. That's because in CSS we cannot target sticky elements (no such pseudo-classes), and we cannot safely add new classes on top of the current HTML structure, lest it cause sudden on-page React issues.

It'd be great if, eventually, our little hack would be re-made into Mintlify's innate feature of the tables :)

Issue #39, opened 2025-09-03
Check that no documentation is lost


verytactical:
- Check every page of old docs has landed somewhere in new docs
- Check that all old pages redirect to relevant new pages — replace all **TODO** and **issue link** destinations with actual routes as they appear. Also, pay attention to the `"/"` destinations, since those often mean that the original page was deleted and no replacement was needed (or not found)
- **(Right before release)** Re-enable the redirects check against the upstream docs.json structure introduced in https://github.com/ton-org/docs/pull/1084

reveloper:
Created old documentation list for matching with new pages: https://docs.google.com/spreadsheets/d/1M-gmnC3nLefCWJBTzHiwMLX4vH2aZYqqCdp4uzQX3Iw/edit?gid=843997444#gid=843997444

reveloper:
Document updated till October 2 commits

PR #1084, opened 2025-11-04
feat(ci): add redirect checks and reconstruct redirects array for all links from the previous TON Docs

### LINKED ISSUE #123: Check redirects on CI
Check that every page that was previously avaiable by URL is either still available, or there is a redirect.

If there is a page that was completely removed, and no possible target where it should land, redirect should land on 404 or main page.

novusnota:
Closes #123 by adding four kinds of redirect checks:

- all sources must be unique
- all destinations must exist (if they are local)
- changes against upstream docs.json structure links must be accounted for
- changes against previous TON Docs links must be accounted for

~~Out of them, **two** are expected to fail — we don't have redirect files for all declared routes AND we don't cover all links from the previous TON Docs. Yet.~~

All checks pass, [see this message](https://github.com/ton-org/docs/pull/1084#issuecomment-3491155216) for details.

### DIFF
--- a/scripts/check-redirects.mjs
+++ b/scripts/check-redirects.mjs
@@ -40,13 +40,15 @@ import {
  */
 
 /**
- * Check that all sources in the redirects array are unique
+ * Check that all sources in the redirects array are unique,
+ * don't point to themselves, and don't override the existing structure paths
  *
  * @param config {Readonly<DocsConfig>} Local docs.json configuration
  * @return {CheckResult}
  */
 const checkUnique = (config) => {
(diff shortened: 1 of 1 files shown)


novusnota:
> Let's add all missing redirects and/or add TODO stubs for destinations and resolve those in #39.

Done!

Instead of explicit "TODO" redirects, which would've led to 404 pages, I've opened new issues and linked to them. And also linked to existing issues when there's no corresponding page stub.

Besides, the "upstream docs.json" check is commented out — it shall be enabled only right before the release. There's a note for that in #39.

verytactical (approved):

PR #1152, opened 2025-11-11
feat: rework stats script

### LINKED ISSUE #1151: Stats script is outdated
It was implemented at the time when there were more stubs than actual pages and when these stubs were unstructured & randomly written. So the page filtering is very bloated and over-agressive.

We can simplify the script and make it more straight forward to align with the current format of the documentation.

Gusarich:
closes #1151


Gusarich:
@codex review

PR #1168, opened 2025-11-11
feat(ci): prohibit adhoc two-letter words in the spell checker

### LINKED ISSUE #1132: CI: further decrease the minimum word length for CSpell checks from 3 to 2
https://cspell.org/docs/Configuration/properties#settings-minwordlength

novusnota:
Closes #1132. From the issue:

> Since the goal is to catch typos in PRs better, we can either manually specify common typos of 2 letter words (and letter omissions in 3 letter words), OR just ban 26*26 two-letter words and then add exceptions (it, be, am, an, etc.).

Well, I started specifying common typos, but then I discovered that the valid range is quite small in comparison, so I just banned `26*26*4` two-letter words instead, excluding valid ones of course.

The number comes from the following pseudo-formula: `[a-z] twice * 4`, where 4 is the number of different spellings a word can have. For example, word `aa` can be written as: `aa`, `aA`, `Aa`, and `AA`. Since the case sensitivity is turned on, each of those options is considered a separate word.

However, I did not set the minimum checked word length to 2 characters in the config due to two reasons:
1. Checks look up the dictionaries, while bans occur regardless of whether a word is allowed in a dictionary, i.e., independently of other checks.
2. Further, those 2-character word dictionaries are lacking, so any unbanned words would then require specifying them in our `resources/dictionaries/custom.txt` file. Not nice.

Overall, **I'm not sure if banning 2k+ words is a good idea.** We might want to only restrict the common letter-omission typos of 3-letter words and be done with it. That said, this might be better in the long run, as people would be more conscious of the spelling errors, and because such typos are tricky to spot with AI. The latter would assume "ar" to be "Augmented Reality" or Argentina's country code rather than a missed "e" in "are".


foundations/addresses/overview.mdx
```diff
@@ -73,7 +73,7 @@ External addresses are closely related to [External messages](/foundations/messa
 
 Actually, external addresses are ignored by the TON Blockchain software altogether, but may be used by external software for its own purposes.
 
-The corresponding Tl-B schemes are as follows:
+The corresponding TL-B schemes are as follows:
```
novusnota, line 76: This was one of the few good catches made with new bans

resources/dictionaries/custom.txt
```diff
@@ -234,6 +234,7 @@ Dodis
 Domínguez
 dotenv
 dton
+Dton
```
anton-trunov: should probably only allow `dTON`

resources/dictionaries/custom.txt
```diff
@@ -234,6 +234,7 @@ Dodis
...
 Duif
```
anton-trunov, line 236: what's Duif? :)

resources/dictionaries/custom.txt
```diff
@@ -648,7 +647,7 @@ Stablecoins
 stakers
 standart
 stdlib
-Ston
+Ston.fi
```
anton-trunov: should only allow Ston.fi or STON.fi

.github/workflows/linter.yml
```diff
@@ -88,8 +88,6 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
```
verytactical, line 92: Is this intentional?

novusnota: Yeah — otherwise the CSpell action took things from the main branch even when the current branch was lagging behind the main. I will add a comment then.

resources/dictionaries/custom.txt
novusnota: This guy: image

resources/dictionaries/custom.txt
novusnota: STON.fi seems favourable, done

.github/workflows/linter.yml
novusnota: Done

resources/dictionaries/custom.txt
novusnota: done

novusnota:
Now that we've fixed `dTON`, I actually think we should explicitly pin the intended spellings for TON-prefixed projects and services. Uno momento...

novusnota:
I've also found a nicer way to ban words that will actually show suggested alternatives — `flagWords`. Hence, the two-letter word ban list can merged with the current deny-list, while the latter contents would fill the `flagWords` array. Replacements in progress...

UPD: That's actually rather tricky to make work without false triggering on style-guide*.mdx files, hmmm

.cspell.jsonc
```diff
@@ -23,43 +30,80 @@
...
+        "AccountChain->accountchain",
+        "BaseChain->basechain",
+        "boc->BoC",
+        "BOC->BoC",
+        "Github->GitHub",
+        "MasterChain->masterchain",
+        "ShardChain->shardchain",
+        "Toncenter->TON Center",
+        "toncoins->Toncoin",
+        "Toncoins->Toncoin",
```
novusnota, line 58: Being pedantic in full accordance with https://ton.org/brand-assets

PR #1238, opened 2025-11-14
feat(foundation/messages): add `StateInit` page

### LINKED ISSUE #1192: [Messages > Deploy message]

Karkarmath:
Closes #1192

### DIFF
--- a/contract-dev/contract-sharding.mdx
+++ b/contract-dev/contract-sharding.mdx
@@ -9,7 +9,7 @@ Some protocols need to store a lot of information in contracts, for example, tok
 
 In such protocols, there is a child contract that initially contains the information identified by a key. In some protocols, it is important to know the Parent contract, which acts as the information manager.
 
-To avoid having to know the key upfront, we do not populate that field in StateInit; we only populate the key field. This makes it easy to locate the required contract later.
+To avoid having to know the key upfront, we do not populate that field in `StateInit`; we only populate the key field. This makes it easy to locate the required contract later.
 
 <Aside
   type="caution"


foundations/messages/state-init.mdx
```diff
@@ -0,0 +1,76 @@
+---
+title: "`StateInit` structure"
+---
+
+`StateInit` is a data structure sent as part of a message, either internal or incoming external, used to
+
+- deploy a smart contract, for which it stores the initial code and data of a contract;
+- [unfreeze](/foundations/status#status-transitions) a contract, where its _last_ code and data.
```
v-lebedev: clarify what exactly is "last code and data"

foundations/messages/state-init.mdx
```diff
@@ -0,0 +1,76 @@
...
+    special:(Maybe TickTock)
+    code:(Maybe ^Cell)
+    data:(Maybe ^Cell)
+    library:(Maybe ^Cell)
+= StateInit;
+```
+
+where the fields are:
+
+- `special`: indicating was this message created by the [tick-tock transaction](/foundations/messages/tick-tock) or not.
```
v-lebedev: afaik, this flag was mean to register a smart contract for tick-tock events, however, it is ignored now

foundations/messages/state-init.mdx
```diff
@@ -0,0 +1,76 @@
...
+- `code`: the initial code of the contract.
+- `data`: the initial data of the contract.
+- `library`: the initial library of the contract.
+- `fixed_prefix_length`: the first `fixed_prefix_length` bits of the address are not compared against the state hash.
```
v-lebedev:
```suggestion
- `fixed_prefix_length`: number of most significant address bits that may differ between the deploy message destination and the address derived from `StateInit`; the remaining bits must match.
```

foundations/messages/state-init.mdx
```diff
@@ -0,0 +1,76 @@
...
+addressFrom(stateInit) = destination
+```
+
+### Example
+
+Assume we have an address `D` starting with `0x5ABC...`. If there are 256 active shardchains, first 8 bits of the address (i.e., `5A`) correspond to shard IDs.
+
+Now, if some other contract `S` is deployed that intends to exchange messages with `D`, and computing an address from its `StateInit` results in `0x2623...`, it will be deployed by default into a shard `26`. Messages from `26` to `5A` will take 2 hops and 2 blocks to arrive at the destination.
+
+To put `S` into the same shard as `D`, `fixed_prefix_length` can be set to 8 in a deploy message sent to `0x2623...` address. Thus, TVM will ignore the discrepancy in first 8 bits of a passed address (`0x2623...`) and computed address (`0x5A23...`), and it will be deployed into the same shard `5A`.
```
v-lebedev:
```suggestion
To put `S` into the same shard as `D`, set `fixed_prefix_length` to 8 and send the deploy message to the address `0x5A23...`. Then TVM will ignore the discrepancy in the first 8 bits between the computed address (`0x2623...`) and the passed destination (`0x5A23...`), because their remaining 248 bits coincide, and `S` will be deployed into shard `5A`.
```

hacker-volodya (changes requested):

PR #1247, opened 2025-11-15
feat(ci): add docs.json bi-directional navigation structure checks

### LINKED ISSUE #167: [FS] Check docs.json against real FS structure
Add CI script that checks:

- there should be no .mdx articles not mentioned in docs.json
- docs.json must not have references to articles that do not exist

novusnota:
Closes #167

### DIFF
--- a/ecosystem/api/toncenter/get-api-key.mdx
+++ b/ecosystem/api/toncenter/get-api-key.mdx
@@ -3,26 +3,45 @@ title: "Get your TON Center API key"
 sidebarTitle: "Get API key"
 ---
 
+import { Image } from '/snippets/image.jsx';
+
 To interact with TON Center’s API at higher rate limits, you’ll need to generate an API key via the official [Telegram bot](https://t.me/toncenter).
 
 ## Open the TON Center bot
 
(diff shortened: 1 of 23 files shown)


tvm/tools/overview.mdx
anton-trunov, line 1: @verytactical says this was intentional

package.json
```diff
@@ -22,13 +23,15 @@
   },
   "dependencies": {
     "cspell": "^9.2.1",
-    "mint": "^4.2.186",
+    "mint": "^4.2.202",
```
anton-trunov: do we need to update `mint` in this PR?

novusnota: Good point, we don't. I've reverted that.

tvm/tools/overview.mdx
novusnota: @grok is this true?

novusnota: It might be the case that this overview was index.mdx in the past. Plus, there's no issue for this file, so it's not really a stub.

verytactical (approved):

tvm/tools/overview.mdx
verytactical: #1257

PR #1302, opened 2025-11-18
feat: start here

### LINKED ISSUE #261: [Start here] Describe structure of documentation
Do not take this ticket until we're done with figuring out top-level structure of the documentation. Can be done at the end of hackathon.

- Describe structure of documentation
    - Basics: must read
    - Step by step: guides for specific common cases
    - Tooling: stuff developer needs
    - Standard contracts: vetted good contracts that already exist
    - Foundations: how TON works
    - Techniques: how to develop new good contracts
- Link to its Github repo

### LINKED ISSUE #263: [Start here] Write the article
Describe minimal set of concepts required to understand rest of the documentation.

Newcomers have a problem understanding what a smart-contract _even is_, and have no idea of the runtime model of blockchain.

verytactical:
Closes #263
Closes #261


start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+| Ecosystem              | Tools and libraries for developing on TON, and how to use them.                              |
+| Payment processing     | Implementation guides for an exchange.                                                       |
+| Standard contracts     | Working with most popular standardized contracts.                                            |
+| Contract development   | Guides on developing smart-contracts.                                                        |
+| Languages              | Reference documentation for TON-specific programming languages.                              |
+| TON Virtual Machine    | Description of the low-level language that runs smart-contracts, and details of the runtime. |
+| Blockchain foundations | Comprehensive description of the blockchain.                                                 |
+| Contribute             | Documentation on writing this documentation.                                                 |
+
+## The minimum
```
anton-trunov: do we need a sub-header here?

image

start-here.mdx
```diff
@@ -1,8 +1,153 @@
 ---
 title: "Start here"
-icon: rocket
```
anton-trunov, line 3: what happened to the rocket?

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
-import { Stub } from '/snippets/stub.jsx';
+import { Aside } from "/snippets/aside.jsx";
 
-<Stub issue="79" />
+This documentation is divided into several sections, with lower-level details appearing later in the list. For example, an existing standard contract may already solve the problem, so there may be no need to implement a new one.
+
+|                        |                                                                                              |
+| ---------------------- | -------------------------------------------------------------------------------------------- |
+| Ecosystem              | Tools and libraries for developing on TON, and how to use them.                              |
+| Payment processing     | Implementation guides for an exchange.                                                       |
```
anton-trunov: - this is too narrow (the overview page for the corresponding section says "Payment processing on TON refers to monitoring and handling blockchain transactions for business applications"
- Needs to be linked to something like https://en.wikipedia.org/wiki/Cryptocurrency_exchange

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+| Standard contracts     | Working with most popular standardized contracts.                                            |
+| Contract development   | Guides on developing smart-contracts.                                                        |
+| Languages              | Reference documentation for TON-specific programming languages.                              |
```
anton-trunov: TL-B is not a programming language

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+| TON Virtual Machine    | Description of the low-level language that runs smart-contracts, and details of the runtime. |
+| Blockchain foundations | Comprehensive description of the blockchain.                                                 |
```
anton-trunov: Perhaps we should highlight the whitepapers here

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+_Mainnet_ is the "real" network. It's where actual payments in Toncoin are made. Applications use mainnet by default, and some don't even have an option to use other networks.
+
+The other network is _testnet_, and it is used by TON developers to check that their applications work correctly before deploying them to mainnet. It uses "test coins" that barely have any value.
+
+Each network is split into _workchains_ that can freely interact with each other, but their implementations may differ significantly.
+
+At the moment, there are two workchains: _basechain_ (`0`) for regular use, and a very similar _masterchain_ (`-1`) for TON's internal [bookkeeping](/foundations/system). The masterchain follows mostly the same rules, except that using it is more expensive to limit the amount of traffic that interferes with TON's internals.
+
+To be freely scalable, each workchain is split into _shards_. The number of shards is [determined dynamically](/foundations/shards) based on the current network load. Internally, every shard is implemented as a separate blockchain. Except for increased latency when messages are sent between distant shards, the effect on the user-facing code is minimal.
```
anton-trunov: "messages" are never explained in the previous paragraphs

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Physically, the blockchain runs on a network of servers, called [nodes](/ecosystem/node/overview). They interact via the _ADNL_ protocol. The physical layer mostly doesn't affect user-facing applications, as they usually use [servers](/ecosystem/api/overview) that proxy JSON HTTP requests into the ADNL network. The official version of such a proxy server is provided by the _liteserver_ software. There are public instances of liteserver, so developers are not required to host one on their own servers.
```
anton-trunov: link to "literserver"

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+## The minimum
+
+<Aside type="tip">
+  This is a condensed description of TON. The rest of the documentation may assume that all of this is already known to the reader.
+</Aside>
+
+_TON_ is a [blockchain](https://en.wikipedia.org/wiki/Blockchain). It provides a distributed platform for hosting applications. Roughly speaking, it works as if it were a single server executing all the code. Such applications are called _smart-contracts_.
+
+The platform runs on a set of servers, called _validators_, owned by individuals or organizations with a large stake in TON and [greatly interested](/ecosystem/staking/overview) in keeping the platform safe, fair, and operational.
```
anton-trunov: "consensus" is never mentioned

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Physically, the blockchain runs on a network of servers, called [nodes](/ecosystem/node/overview). They interact via the _ADNL_ protocol. The physical layer mostly doesn't affect user-facing applications, as they usually use [servers](/ecosystem/api/overview) that proxy JSON HTTP requests into the ADNL network. The official version of such a proxy server is provided by the _liteserver_ software. There are public instances of liteserver, so developers are not required to host one on their own servers.
+
+ADNL works over [TCP](https://en.wikipedia.org/wiki/Transmission_Control_Protocol), so it's impossible to connect to ADNL from a web page. Browsers [intentionally](https://security.stackexchange.com/q/100827) do not provide an API to send arbitrary TCP traffic. Servers do not have this limitation, as they do not use a browser and can interact with the blockchain over either ADNL or HTTP.
```
anton-trunov: let's provide a better link than a link to stachexchange (those answers can be edited by the authos or moderators)

anton-trunov: "Servers do not have this limitation, as they do not use a browser", let's remove "as they do not use a browser"

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+![Statuses overview](/resources/images/states.png)
+
+It's easiest to visualize the blockchain as a set of [accounts](/foundations/addresses/overview). Each account has an address and a status.
+
+Over its lifetime, an account changes its [status](/foundations/status) among four values:
+
+- `nonexist`: There wasn't a single operation with the account, or it was removed. It has neither a balance, nor code.
+- `uninit`: If some Toncoin is transferred to an account, it now exists, but there is still no smart contract code on it. It now has a balance.
+- `active`: After a deploy message (see below) with code and initial data is sent to an account, it becomes active and can process other messages. It now has a balance, code, and internal state.
+- `frozen`: If an account is overdue on its storage fees, it will be frozen until the fees are paid. If the balance becomes too small, the account goes completely bankrupt, is removed, and ceases to exist.
```
anton-trunov: link to "storeage fees"

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+The _code_ on an active account is a smart contract. The term _"contract"_ is often used for an account that holds the code.
+
+[Internal address](/foundations/addresses/overview#internal-addresses) of an account is a 256-bit number and is usually [formatted](/foundations/addresses/formats) like this: `EQDKbjIcfM6ezt8KjKJJLshZJJSqX7XOA4ff-W72r5gqPrHF`.
+
+There are also [external addresses](/foundations/addresses/overview#external-addresses) for things outside the blockchain. In any situation where an external address is involved, a special "none" address may be provided.
```
anton-trunov: can we link to "none" address?

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Most development is done in the [Tolk](/languages/tolk/overview), a high-level programming language. Its compiler is included in the [Blueprint](/contract-dev/blueprint/overview) development environment and will be available to all projects created from the template.
+
+Originally, [Fift](/languages/fift/overview), a Forth-like assembly language, and [FunC](/languages/func/overview), a C-like intermediate-level language, were used for contract development.
+
+When execution [starts](/tvm/initialization), the message and current account state are provided to the code. By the end of execution, the destination account might change its state or code, or send internal or outgoing external messages.
+
+The execution follows a procedure whose steps are called [phases](/foundations/phases). [Fees](/foundations/fees) are deducted during this process. Fees might be deducted from the account's balance or from the message, depending on the [mode](/foundations/messages/modes) of the message.
```
anton-trunov: Unless I'm mistaken, not only the message mode, but the contract also can influence how fees are paid

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+If something goes wrong, a non-zero [exit code](/tvm/exit-codes) might be returned, no changes to state or code are saved, and no further messages are sent.
+
+![Trace overview](/resources/images/trace.png)
+
+The only reason any code can be executed is when some account has received a message. Internal messages can only be sent by another contract executing some code, and that contract must have received a message from somewhere too. In the end, every message can be considered part of some _trace_: a tree of messages between accounts that starts with an incoming external message, continues with internal messages, and possibly ends with some outgoing external messages.
+
+If an external service needs to extract data from the blockchain, it can call a [get method](/tvm/get-method): a handler of incoming external messages in the code of some account. It's called a "method" because it acts like one: the message provides its parameters, and its execution returns a value. Get methods cannot make any changes to the blockchain.
```
anton-trunov: Let's also state explicitly that contracts cannot call each other's get methods: it is a too common question

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+The other source is a [wallet](/standard/wallets/how-it-works). A wallet is an account with a specific kind of smart contract deployed on it, which can handle _transfer_ incoming external messages: requests to send an internal message to another account. There are several [implementations](/standard/wallets/comparison) of wallets, with varying extra functionality and protection. [`V5R1`](/standard/wallets/v5) is the latest official general-purpose wallet. The text below describes only the functionality common to most wallets.
```
anton-trunov: something with the grammar here: "transfer incoming external messages"

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+A transfer message consists of a destination address and, optionally, an internal message, both serialized and signed with a private key. The wallet stores a public key and uses it to check that a transfer message was signed with the corresponding private key. If the check succeeds, it sends that internal message to the destination address. This ensures only the user (or a service) who knows the private key can use the wallet.
+
+Private and public [keys](https://en.wikipedia.org/wiki/Public-key_cryptography) are generated in a program or service outside the blockchain and determine the address of the wallet account. Their text representation is called a [mnemonic](/standard/wallets/mnemonics).
```
anton-trunov: maybe just this link should be used in the previous paragraph

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+There are two copies of TON blockchain: mainnet and testnet.
+
+_Mainnet_ is the "real" network. It's where actual payments in Toncoin are made. Applications use mainnet by default, and some don't even have an option to use other networks.
+
+The other network is _testnet_, and it is used by TON developers to check that their applications work correctly before deploying them to mainnet. It uses "test coins" that barely have any value.
+
+Each network is split into _workchains_ that can freely interact with each other, but their implementations may differ significantly.
+
+At the moment, there are two workchains: _basechain_ (`0`) for regular use, and a very similar _masterchain_ (`-1`) for TON's internal [bookkeeping](/foundations/system). The masterchain follows mostly the same rules, except that using it is more expensive to limit the amount of traffic that interferes with TON's internals.
```
Shvandre: > except that using it is more expensive to limit the amount of traffic that interferes with TON's internals.

except that using it is more expensive in order to limit traffic that interferes with TON’s internal processes.

start-here.mdx
Shvandre: what about linking https://en.wikipedia.org/wiki/Shard_(database_architecture)?

start-here.mdx
Shvandre: But liteservers do not expose http api by default.....

Also the whole paragraph (I mean line 44) is unclear, need more context here

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+The platform runs on a set of servers, called _validators_, owned by individuals or organizations with a large stake in TON and [greatly interested](/ecosystem/staking/overview) in keeping the platform safe, fair, and operational.
+
+_Toncoin_ is TON's primary [cryptocurrency](https://en.wikipedia.org/wiki/Cryptocurrency). It is used to pay for the execution of smart contracts, the storage of their data, and network traffic. Such payments are called [fees](/foundations/fees).
+
+There are two copies of TON blockchain: mainnet and testnet.
+
+_Mainnet_ is the "real" network. It's where actual payments in Toncoin are made. Applications use mainnet by default, and some don't even have an option to use other networks.
+
+The other network is _testnet_, and it is used by TON developers to check that their applications work correctly before deploying them to mainnet. It uses "test coins" that barely have any value.
```
Shvandre, line 37:
```suggestion
The other network is _testnet_, and it is used by TON developers to check that their applications work correctly before deploying them to mainnet. It uses "test coins" that [barely have any value](/contract-dev/testing/testnet-tokens).
```

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+It's easiest to visualize the blockchain as a set of [accounts](/foundations/addresses/overview). Each account has an address and a status.
+
+Over its lifetime, an account changes its [status](/foundations/status) among four values:
+
+- `nonexist`: There wasn't a single operation with the account, or it was removed. It has neither a balance, nor code.
+- `uninit`: If some Toncoin is transferred to an account, it now exists, but there is still no smart contract code on it. It now has a balance.
+- `active`: After a deploy message (see below) with code and initial data is sent to an account, it becomes active and can process other messages. It now has a balance, code, and internal state.
+- `frozen`: If an account is overdue on its storage fees, it will be frozen until the fees are paid. If the balance becomes too small, the account goes completely bankrupt, is removed, and ceases to exist.
+
+The _code_ on an active account is a smart contract. The term _"contract"_ is often used for an account that holds the code.
```
Shvandre: think about bold text here, as now "contract" intersects with "is"
image

start-here.mdx
Shvandre: I'd skip that part about external addresses, it's too hard and not everyone needs to know it

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+[Internal address](/foundations/addresses/overview#internal-addresses) of an account is a 256-bit number and is usually [formatted](/foundations/addresses/formats) like this: `EQDKbjIcfM6ezt8KjKJJLshZJJSqX7XOA4ff-W72r5gqPrHF`.
+
+There are also [external addresses](/foundations/addresses/overview#external-addresses) for things outside the blockchain. In any situation where an external address is involved, a special "none" address may be provided.
+
+![Blockchain overview](/resources/images/blockchain.png)
```
Shvandre: Please, we need dark themed pictures, my eyes are bleeding

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Addresses specify where [messages](/foundations/messages/overview) should be delivered. There are three types of messages:
+
+- [internal](/foundations/messages/internal) messages are sent between accounts;
+- [incoming external](/foundations/messages/external-in) messages are sent from code outside the blockchain to a contract;
+- [outgoing external](/foundations/messages/external-out) messages are broadcast to the external network and can be read by anyone.
+
+Every internal message should have some Toncoin attached to it so that it can pay for the cost of handling it. External messages cannot have Toncoin attached to them because they spend some of their time outside the blockchain, and only the blockchain can prove the transfer of Toncoin is correct.
```
Shvandre, line 82: > External messages cannot have Toncoin attached to them because they spend some of their time outside the blockchain, and only the blockchain can prove the transfer of Toncoin is correct.

I got your idea but it's better co remove this clarification

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Incoming external messages come from an external address, and outgoing external messages go to an external address.
+
+The _state_ of the account changes only when it handles messages. Messages also change the account's balance. To have a state and a balance, and to handle messages, the account must be `active`.
+
+Formally, a message is only an intent: it has a destination, possibly some Toncoin, and data. After the message is handled and all the necessary changes are made to the blockchain, the message is packed, along with a description of those changes, into a single packet of data, [transaction](/foundations/messages/ordinary-tx).
```
Shvandre: I'd better link /messages/overview#transactions

I wouldn't say that the message is "packed". It's better to say, that "Transactions reflect the state changes on accounts"

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+A message might also be a [deploy message](/foundations/messages/deploy) if it has contract code and its initial state (`StateInit`) attached. When such a message is sent, the hash of the code and initial state must match the destination address. If they do, the code and state are stored in the account at that address, and the account becomes active. Both the code and state stored in the account may change in the future, but its address will remain the same as when it was originally deployed.
+
+Internal and incoming external messages execute the account's code. The code is interpreted by [TON Virtual Machine](/tvm/overview) (TVM). It is written in _bitcode_, a binary format specific to TVM. In the future, TVM might support multiple binary languages, _codepages_, but at the moment there is only codepage 0 (`CP0`).
+
+Most development is done in the [Tolk](/languages/tolk/overview), a high-level programming language. Its compiler is included in the [Blueprint](/contract-dev/blueprint/overview) development environment and will be available to all projects created from the template.
```
Shvandre, line 94: It hurts when you don't see Tact here. Let's mention it somehow

start-here.mdx
Shvandre: > a handler of incoming external messages in the code of some account.

get method is not an incoming external

start-here.mdx
Shvandre: The other source of what?

start-here.mdx
Shvandre: Formally key is derived from mnemonic, it's one way process, idk whether we should mention it or not

start-here.mdx
gabriel-fallen: As far as I remember, the last argument about "what is a programming language" didn't reach a consensus. 😉 

Thus, I would rather not draw attention to this debatable distinction in the documentation, aimed at _other_ topics. 😁

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+| Contract development   | Guides on developing smart-contracts.                                                        |
+| Languages              | Reference documentation for TON-specific programming languages.                              |
+| TON Virtual Machine    | Description of the low-level language that runs smart-contracts, and details of the runtime. |
+| Blockchain foundations | Comprehensive description of the blockchain.                                                 |
+| Contribute             | Documentation on writing this documentation.                                                 |
+
+## The minimum
+
+<Aside type="tip">
+  This is a condensed description of TON. The rest of the documentation may assume that all of this is already known to the reader.
```
gabriel-fallen, line 24:
```suggestion
  **Read this first!** This is a condensed description of TON. The rest of the documentation may assume that all of this is already known to the reader.
```
"The rest of the documentation may assume that all of this is already known to the reader" seems very important to me (as a newcomer), so I think it deserves more visibility and emphasis. Maybe also promote "tip" to "warning" or something?

start-here.mdx
verytactical: Unfortunately not.

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+</Aside>
+
+_TON_ is a [blockchain](https://en.wikipedia.org/wiki/Blockchain). It provides a distributed platform for hosting applications. Roughly speaking, it works as if it were a single server executing all the code. Such applications are called _smart-contracts_.
```
Gusarich: "distributed platform for hosting applications", then immediately "as if it were a single server executing all the code"

I get the point but it can be misleading for newcomers

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+The platform runs on a set of servers, called _validators_, owned by individuals or organizations with a large stake in TON and [greatly interested](/ecosystem/staking/overview) in keeping the platform safe, fair, and operational.
+
+_Toncoin_ is TON's primary [cryptocurrency](https://en.wikipedia.org/wiki/Cryptocurrency). It is used to pay for the execution of smart contracts, the storage of their data, and network traffic. Such payments are called [fees](/foundations/fees).
+
+There are two copies of TON blockchain: mainnet and testnet.
```
Gusarich: I wouldn't call these "copies". Mainnet and testnet don't have much in common as chains, apart from the structure.

start-here.mdx
Gusarich: "the effect on the user-facing code is minimal." - I'd mention cross-contract race conditions here (do we have a separate article for that?)

start-here.mdx
gabriel-fallen: You have to read the Wikipedia page _very closely_ to find anything on application hosting. Thus, the first two sentences appear disconnected. I suggest building more of a bridge along the lines of

"It provides a distributed platform for storing data and code, as well as running computations, all the ingredients to host applications."

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Physically, the blockchain runs on a network of servers, called [nodes](/ecosystem/node/overview). They interact via the _ADNL_ protocol. The physical layer mostly doesn't affect user-facing applications, as they usually use [servers](/ecosystem/api/overview) that proxy JSON HTTP requests into the ADNL network. The official version of such a proxy server is provided by the _liteserver_ software. There are public instances of liteserver, so developers are not required to host one on their own servers.
+
+ADNL works over [TCP](https://en.wikipedia.org/wiki/Transmission_Control_Protocol), so it's impossible to connect to ADNL from a web page. Browsers [intentionally](https://security.stackexchange.com/q/100827) do not provide an API to send arbitrary TCP traffic. Servers do not have this limitation, as they do not use a browser and can interact with the blockchain over either ADNL or HTTP.
+
+![Statuses overview](/resources/images/states.png)
```
Gusarich: why not put this image below? when reading from top to bottom it's not clear what this image is related to from the parts above..

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+Over its lifetime, an account changes its [status](/foundations/status) among four values:
+
+- `nonexist`: There wasn't a single operation with the account, or it was removed. It has neither a balance, nor code.
+- `uninit`: If some Toncoin is transferred to an account, it now exists, but there is still no smart contract code on it. It now has a balance.
+- `active`: After a deploy message (see below) with code and initial data is sent to an account, it becomes active and can process other messages. It now has a balance, code, and internal state.
+- `frozen`: If an account is overdue on its storage fees, it will be frozen until the fees are paid. If the balance becomes too small, the account goes completely bankrupt, is removed, and ceases to exist.
+
+The _code_ on an active account is a smart contract. The term _"contract"_ is often used for an account that holds the code.
+
+[Internal address](/foundations/addresses/overview#internal-addresses) of an account is a 256-bit number and is usually [formatted](/foundations/addresses/formats) like this: `EQDKbjIcfM6ezt8KjKJJLshZJJSqX7XOA4ff-W72r5gqPrHF`.
```
Gusarich: I'd mention address formats here just for clarity. Because devs will sometimes see other formats too..

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+There are also [external addresses](/foundations/addresses/overview#external-addresses) for things outside the blockchain. In any situation where an external address is involved, a special "none" address may be provided.
+
+![Blockchain overview](/resources/images/blockchain.png)
+
+Addresses specify where [messages](/foundations/messages/overview) should be delivered. There are three types of messages:
+
+- [internal](/foundations/messages/internal) messages are sent between accounts;
+- [incoming external](/foundations/messages/external-in) messages are sent from code outside the blockchain to a contract;
+- [outgoing external](/foundations/messages/external-out) messages are broadcast to the external network and can be read by anyone.
```
Gusarich: all messages can be "read by anyone".. it's not limited to external-out messages, so I'd remove this phrase to prevent misunderstandings

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Every internal message should have some Toncoin attached to it so that it can pay for the cost of handling it. External messages cannot have Toncoin attached to them because they spend some of their time outside the blockchain, and only the blockchain can prove the transfer of Toncoin is correct.
+
+Incoming external messages come from an external address, and outgoing external messages go to an external address.
+
+The _state_ of the account changes only when it handles messages. Messages also change the account's balance. To have a state and a balance, and to handle messages, the account must be `active`.
```
Gusarich: transaction can be triggered without a message.. for example, tick-tock and due-payment transactions (this one is applicable to "regular" smart contracts that devs write)..

start-here.mdx
Gusarich: "or from the message" - what does that mean?

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Originally, [Fift](/languages/fift/overview), a Forth-like assembly language, and [FunC](/languages/func/overview), a C-like intermediate-level language, were used for contract development.
+
+When execution [starts](/tvm/initialization), the message and current account state are provided to the code. By the end of execution, the destination account might change its state or code, or send internal or outgoing external messages.
+
+The execution follows a procedure whose steps are called [phases](/foundations/phases). [Fees](/foundations/fees) are deducted during this process. Fees might be deducted from the account's balance or from the message, depending on the [mode](/foundations/messages/modes) of the message.
+
+Execution cost is first measured in _gas_ units, then converted to Toncoin. This unit is separate so that if code execution becomes computationally cheaper (or more expensive), validators can vote to change the [price of gas](/foundations/config#param-20-and-21:-gas-prices) in Toncoin.
+
+If something goes wrong, a non-zero [exit code](/tvm/exit-codes) might be returned, no changes to state or code are saved, and no further messages are sent.
```
Gusarich: there's not a single mention of bounces on this page... and i think in this specific place it would be a nice thing to mention because it's very important!

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+![Trace overview](/resources/images/trace.png)
+
+The only reason any code can be executed is when some account has received a message. Internal messages can only be sent by another contract executing some code, and that contract must have received a message from somewhere too. In the end, every message can be considered part of some _trace_: a tree of messages between accounts that starts with an incoming external message, continues with internal messages, and possibly ends with some outgoing external messages.
```
Gusarich: "The only reason any code can be executed is when some account has received a message." - again, same as one of the above comments..

Gusarich: "and possibly ends with some outgoing external messages." - external-out messages aren't necessarily in the end, so it's misleading

start-here.mdx
gabriel-fallen:
```suggestion
The platform runs on a set of servers, called _validators_, owned by individuals or organizations with a large stake in TON and [great interest](/ecosystem/staking/overview) in keeping the platform safe, fair, and operational.
```

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+The other source is a [wallet](/standard/wallets/how-it-works). A wallet is an account with a specific kind of smart contract deployed on it, which can handle _transfer_ incoming external messages: requests to send an internal message to another account. There are several [implementations](/standard/wallets/comparison) of wallets, with varying extra functionality and protection. [`V5R1`](/standard/wallets/v5) is the latest official general-purpose wallet. The text below describes only the functionality common to most wallets.
+
+A transfer message consists of a destination address and, optionally, an internal message, both serialized and signed with a private key. The wallet stores a public key and uses it to check that a transfer message was signed with the corresponding private key. If the check succeeds, it sends that internal message to the destination address. This ensures only the user (or a service) who knows the private key can use the wallet.
```
Gusarich: "A transfer message consists of ..." this is not clear which "transfer message" is meant here - the one sent to the wallet? or the one sent by wallet? let's call it an "external message accepted by a wallet" or something like that

start-here.mdx
Gusarich: "and determine the address of the wallet account." - address calculation was mentioned above on the page, so this part can sound like wallets have some _special_ way of determining addresses... I'd make it more clear here just by noting that public key is in the stateinit of the wallet

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Private and public [keys](https://en.wikipedia.org/wiki/Public-key_cryptography) are generated in a program or service outside the blockchain and determine the address of the wallet account. Their text representation is called a [mnemonic](/standard/wallets/mnemonics).
+
+Before the code of the wallet can be deployed on an account with an incoming external message, some other account has to transfer Toncoin to the wallet account. External messages cannot have any Toncoin attached to them, and otherwise there wouldn't be enough funds to pay fees to handle the message that deploys a wallet. This transfer to a new wallet account usually comes from an exchange or another user. In testnet, there is [a bot](/ecosystem/wallet-apps/web#add-funds-into-the-wallet) that sends test coins to an account for free.
```
Gusarich: "External messages cannot ..., and otherwise ..."

I think you supposed the "otherwise" part to be related to the previous sentence, but it's not clear from the way it's written right now

start-here.mdx
verytactical: It should be linked from the `/foundations/shards` article.

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Usually, a transfer message is sent without an internal message and only instructs the wallet to transfer some Toncoin to a destination address. This is the reason this type of smart contract is called a wallet.
```
Gusarich: "Usually, a transfer message is sent without an internal message" - what does this mean?

start-here.mdx
verytactical: Thanks, I needed this link elsewhere, but this is the wrong place to link it.

start-here.mdx
verytactical: We need pictures that are not drawn literally in **Pain**t.

verytactical: #262

start-here.mdx
verytactical: No.

start-here.mdx
verytactical: That section of documentation is poorly written and doesn't even link to the article about transactions.

The wording should indeed be improved here, thanks for the suggestion.

start-here.mdx
verytactical: It's sneakily mentioned below.

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
 
-import { Stub } from '/snippets/stub.jsx';
+import { Aside } from "/snippets/aside.jsx";
 
-<Stub issue="79" />
+This documentation is divided into several sections, with lower-level details appearing later in the list. For example, an existing standard contract may already solve the problem, so there may be no need to implement a new one.
+
+|                        |                                                                                              |
+| ---------------------- | -------------------------------------------------------------------------------------------- |
+| Ecosystem              | Tools and libraries for developing on TON, and how to use them.                              |
```
Kaladin13: cross-links to referenced sections here

start-here.mdx
Kaladin13: TON is not a blockchain, it's a network, blockchain is a part of it. There exists whitepaper that's called "The Open Network Blockchain", further implying this statement

start-here.mdx
Kaladin13: Agreed, copies is a bad word here

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+_TON_ is a [blockchain](https://en.wikipedia.org/wiki/Blockchain). It provides a distributed platform for hosting applications. Roughly speaking, it works as if it were a single server executing all the code. Such applications are called _smart-contracts_.
+
+The platform runs on a set of servers, called _validators_, owned by individuals or organizations with a large stake in TON and [greatly interested](/ecosystem/staking/overview) in keeping the platform safe, fair, and operational.
+
+_Toncoin_ is TON's primary [cryptocurrency](https://en.wikipedia.org/wiki/Cryptocurrency). It is used to pay for the execution of smart contracts, the storage of their data, and network traffic. Such payments are called [fees](/foundations/fees).
+
+There are two copies of TON blockchain: mainnet and testnet.
+
+_Mainnet_ is the "real" network. It's where actual payments in Toncoin are made. Applications use mainnet by default, and some don't even have an option to use other networks.
```
Kaladin13: > Applications use mainnet by default, and some don't even have an option to use other networks.

I don't understand what does that mean

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+The other network is _testnet_, and it is used by TON developers to check that their applications work correctly before deploying them to mainnet. It uses "test coins" that barely have any value.
+
+Each network is split into _workchains_ that can freely interact with each other, but their implementations may differ significantly.
```
Kaladin13: Workchain article cross-link

start-here.mdx
Kaladin13: From the article flow and context it does't make any sense why we have number in brackets next to workchain names. Either describe workchain id or remove them

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Physically, the blockchain runs on a network of servers, called [nodes](/ecosystem/node/overview). They interact via the _ADNL_ protocol. The physical layer mostly doesn't affect user-facing applications, as they usually use [servers](/ecosystem/api/overview) that proxy JSON HTTP requests into the ADNL network. The official version of such a proxy server is provided by the _liteserver_ software. There are public instances of liteserver, so developers are not required to host one on their own servers.
+
+ADNL works over [TCP](https://en.wikipedia.org/wiki/Transmission_Control_Protocol), so it's impossible to connect to ADNL from a web page. Browsers [intentionally](https://security.stackexchange.com/q/100827) do not provide an API to send arbitrary TCP traffic. Servers do not have this limitation, as they do not use a browser and can interact with the blockchain over either ADNL or HTTP.
+
+![Statuses overview](/resources/images/states.png)
+
+It's easiest to visualize the blockchain as a set of [accounts](/foundations/addresses/overview). Each account has an address and a status.
```
Kaladin13, line 51: We need separate heading or some intro into this section, it's starts very unexpected

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Addresses specify where [messages](/foundations/messages/overview) should be delivered. There are three types of messages:
+
+- [internal](/foundations/messages/internal) messages are sent between accounts;
+- [incoming external](/foundations/messages/external-in) messages are sent from code outside the blockchain to a contract;
+- [outgoing external](/foundations/messages/external-out) messages are broadcast to the external network and can be read by anyone.
+
+Every internal message should have some Toncoin attached to it so that it can pay for the cost of handling it. External messages cannot have Toncoin attached to them because they spend some of their time outside the blockchain, and only the blockchain can prove the transfer of Toncoin is correct.
+
+Incoming external messages come from an external address, and outgoing external messages go to an external address.
```
Kaladin13, line 84: Mention that external-outs are like blockchain logs, events in other blockchains

start-here.mdx
Kaladin13:
```suggestion
The _state_ of the account changes only when it handles messages. Messages also change the account's balance. To have a state and a balance, and to handle messages, the account must be `active`. This state change is called "transaction"
```

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+- use [an SDK](/ecosystem/sdks) that simplifies working with an API by wrapping its methods in a more user-friendly interface; the most popular TypeScript SDK for this is `@ton/ton`.
+
+When an application has to use a wallet to prove the user's identity, it needs access to the wallet's private key. Giving arbitrary applications access to the private key is insecure, as they could perform arbitrary actions with the wallet. To address this, there is the [TON Connect](/ecosystem/ton-connect/overview) set of SDKs that provide interfaces
+
+- for an application to perform an action through the wallet app;
+- for a wallet app to handle these requests.
+
+For example, Telegram apps use TON Connect to access a wallet that is integrated into Telegram.
+
+All data on TON is stored as trees of [cells](/foundations/serialization/cells): the state of contracts, their code, and messages. Each cell stores up to 1023 bits of data and can have up to 4 _refs_ to other cells. Each cell has a hash computed from its bits and refs. Because the hash of a cell that directly or indirectly references itself would require knowing the same hash, creating cyclic data structures is impossible.
```
Kaladin13, line 157: New heading or some explanation, very rough change of theme, from connect to cells

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Standard serialization of such a data structure into a single binary string is the [bag of cells](/foundations/serialization/boc) (_BoC_). Compilers of smart-contract programming languages produce a BoC with code. When there is a need to store cells in a file or send them over the network, they are usually serialized into a BoC.
+
+To tell other developers how a certain type of data is stored in cells, a [TL-B](/languages/tl-b/overview) schema language is used. Its purpose is similar to that of [protocol buffers](https://en.wikipedia.org/wiki/Protocol_Buffers) or [binary templates](https://en.wikipedia.org/wiki/010_Editor#Binary_Templates), but it provides more features for structuring data at the bit level.
+
+There are libraries for assembling data structures out of cells. For TypeScript, the most popular one is [`@ton/core`](https://github.com/ton-org/ton-core).
```
Kaladin13: SDK page cross link

start-here.mdx
verytactical: This was the intent behind making an `<Aside>` here. I raised its level from "tip" to "caution" instead of modifying the text.

start-here.mdx
verytactical: Very nice wording. Thanks!

start-here.mdx
verytactical: There is a mention below

```
it is important to keep in mind because it might create [race conditions](/contract-dev/security#race-condition-of-messages).
```

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+Incoming external messages come from an external address, and outgoing external messages go to an external address.
+
+The _state_ of the account changes only when it handles messages. Messages also change the account's balance. To have a state and a balance, and to handle messages, the account must be `active`.
+
+Formally, a message is only an intent: it has a destination, possibly some Toncoin, and data. After the message is handled and all the necessary changes are made to the blockchain, the message is packed, along with a description of those changes, into a single packet of data, [transaction](/foundations/messages/ordinary-tx).
+
+A message might also be a [deploy message](/foundations/messages/deploy) if it has contract code and its initial state (`StateInit`) attached. When such a message is sent, the hash of the code and initial state must match the destination address. If they do, the code and state are stored in the account at that address, and the account becomes active. Both the code and state stored in the account may change in the future, but its address will remain the same as when it was originally deployed.
```
gabriel-fallen:
```suggestion
A message might also be a [deploy message](/foundations/messages/deploy) if it has contract code and its initial state (`StateInit`) attached. When such a message is sent, the hash taken of both the code and initial state must match the destination address. If it does, the code and state are stored in the account at that address, and the account becomes active. Both the code and state stored in the account may change in the future, but its address will remain the same as when it was originally deployed.
```

start-here.mdx
Karkarmath: > The platform runs on a set of servers, called _validators_...

At the same time, on line 44 the following is written:

> the blockchain runs on a network of servers, called nodes...

It should be consistent.

start-here.mdx
Karkarmath: Uninit and frozen accounts also have the state and balance.

start-here.mdx
Karkarmath: > A message might also be a [deploy message](/foundations/messages/deploy) if it has contract code and its initial state (`StateInit`) attached.

A code is a part of account's initial state.

start-here.mdx
verytactical: We've changed the wording already, based on feedback above

start-here.mdx
gabriel-fallen:
```suggestion
If something goes wrong, a non-zero [exit code](/tvm/exit-codes) might be returned, no changes to state or code are saved, and no messages are sent.
```

The original version creates an impression that some messages (before an error occurred) are sent, but after that point, no _further_ messages are sent. I guess that's not the case, as that would break the transactional semantics.

start-here.mdx
verytactical: Most of them don't have an overview page to link to, and they're already in the navigation on the left.

start-here.mdx
verytactical: We barely scratch the part where it is a network in the whole documentation. For the majority of the users (well, almost everyone except for Nikolai and xssnick) it is a blockchain. We can come back to this in the future.

start-here.mdx
```diff
@@ -1,8 +1,153 @@
 ---
 title: "Start here"
-icon: rocket
 ---
 
-import { Stub } from '/snippets/stub.jsx';
+import { Aside } from "/snippets/aside.jsx";
 
-<Stub issue="79" />
+This documentation is divided into several sections, with lower-level details appearing later in the list. For example, an existing standard contract may already solve the problem, so there may be no need to implement a new one.
```
jeshecdom, line 8: I don't see how this sentence:
 
> For example, an existing standard contract may already solve the problem, so there may be no need to implement a new one.

follows logically from the previous sentence:

> This documentation is divided into several sections, with lower-level details appearing later in the list.

The "documentation" sentence is describing the list, but then why the example talks about "no need to implement a new one [contract]"? 

Perhaps the intention was to say something like the following?.... I am just trying to guess what was the intention:

"For example, an existing standard contract may already solve the problem, so there is no need to read lower level sections in the list in order to re-implement such contract."

start-here.mdx
jeshecdom, line 37: > "test coins" that barely have any value.

So, do they have some value in some situations, like paying for stuff beyond testing? Or what does "value" mean here?

start-here.mdx
verytactical: We cannot really put headings here, as it describes a graph of terms, depending on each other. At the moment, we don't even know the full graph.

start-here.mdx
jeshecdom: > as they usually use [servers]

Is "they" here referring to the user-facing applications or to the nodes?

> so developers are not required to host one on their own servers.

"servers" here is referring to the nodes or to the liteserver proxies?

Sorry, the paragraph at line 44 confused me.

Karkarmath:
To be honest, it was a bit hard to keep attention and be concentrated until the end of the article. So, it is a really hard article for the newcomers.

At the same time, it is a great article describing almost all aspects of the Ton blockchain. I think as a first approximation it is quite good, but need to be adapted later.

start-here.mdx
jeshecdom: > as they do not use a browser and can interact with the blockchain over either ADNL or HTTP.

The entire paragraph is talking about TCP and ADNL. But then, HTTP makes an appearance at the end. Why the sudden change in focus from TCP to HTTP? Is it because of the mention of web pages and browsers? Also, why is it important to emphasize browsers and web pages here, if the paragraph is talking about communication between servers?

start-here.mdx
jeshecdom: > If an account is overdue on its storage fees, it will be frozen until the fees are paid. If the balance becomes too small, the account goes completely bankrupt, is removed, and ceases to exist.

It is not clear when an account ceases to exist compared to its frozen state. The paragraph says that when the balance is "too small", the account bankrupts and ceases to exist, but then, what does it mean to be overdue on storage fees? If there is not enough balance to pay the storage fees doesn't it mean that the contract will jump immediately from its frozen state to nonexistent?

start-here.mdx
jeshecdom: The image has a dark spot to the right, as if it was cut. At first I thought that my browser was not showing me the complete image, so I clicked on it =P

Also, at first I thought that the label "<- external messages ->" referred to the boundary oval and got confused, but then I realized that it was referring to the arrows coming in and out of the oval. Maybe either place the "external messages" label outside the oval, or increase the length of the "<-   ->" in the label so that they almost touch the arrows depicting the external messages.

start-here.mdx
verytactical: This was about `OutList`. This would require a description of the list, actions, the send message action, and computation phases.

I would rather prefer to leave this as is, because "no message" would definitely be read incorrectly.

resources/images/blockchain.png
novusnota, line 1: Perhaps, this should be replaced with an svg with the same contents — then, it'll work in dark themes too.

Otherwise, please add an image for a dark theme

resources/images/trace.png
novusnota, line 1: On the contrary, this needs an image for a light theme

resources/images/blockchain.png
verytactical: Please read other comments before adding new comments :)

start-here.mdx
```diff
@@ -1,8 +1,153 @@
...
+
+![Statuses overview](/resources/images/states.png)
+
+It's easiest to visualize the blockchain as a set of [accounts](/foundations/addresses/overview). Each account has an address and a status.
+
+Over its lifetime, an account changes its [status](/foundations/status) among four values:
+
+- `nonexist`: There wasn't a single operation with the account, or it was removed. It has neither a balance, nor code.
+- `uninit`: If some Toncoin is transferred to an account, it now exists, but there is still no smart contract code on it. It now has a balance.
+- `active`: After a deploy message (see below) with code and initial data is sent to an account, it becomes active and can process other messages. It now has a balance, code, and internal state.
```
jeshecdom, line 63: > After a deploy message (see below).

The image actually occurs above these paragraphs.

resources/images/blockchain.png
novusnota: I've read all comments and there were no mention of image themes except for the `<Image />` component use (requested in AI review)

start-here.mdx
jeshecdom: The paragraph starts mentioning "procedure":

> .... execution follows a procedure whose steps....

Then it changes to "process" in the next sentence:

> Fees are deducted during this process.

Also, it is unclear if fees are deducted at each phase or only at the end.

start-here.mdx
verytactical: There already exists a standard contract, and there is no need to implement a new one, literally.

Issue #1366, opened 2025-11-20
Workflow


verytactical:
Description of our processes.

Do not discuss anything in comments, if you have write access.

> [!NOTE]
> Many issues have recently been closed as "not planned," not because they were truly not planned, but because they had become stale, which made the current task scope too broad and unwieldy. Once there is enough time to return to "not planned" tasks, it would be great to do so.
>
> Given this, please always search for existing issues, including closed issues, before opening a new one. Even if they are closed, reopen them and proceed with the work.
>
> All issues closed as "not planned" MUST STAY in the [tree-like issue Backlog](https://github.com/ton-org/docs/issues/5).

verytactical:
## Create an issue

When the change is small, well-discussed in advance, or comes from external contributors, one may skip creation of an issue and proceed straight to the PR. That is assuming the contribution is not controversial and is understood by the person behind the submission in its entirety.

### Find an appropriate place in the [Backlog](https://github.com/ton-org/docs/issues/5).

- `Top-level articles` is for articles on the top level.
- `CI and processes` is for work on the repo, checks, and workflow.
- `Styleguide` is for the styleguide.
- `Structure` is for global structural edits, and it's very likely you're in the wrong place.
- The rest are for corresponding sections of docs.
  - The child `Plan`s follow the structure of documentation. If they don't, fix it!

### Press the "Create sub-issue" button.

<img width="454" height="162" alt="Image" src="https://github.com/user-attachments/assets/32c10871-1967-46b5-b178-7e4607b9ed3d" />

**DO NOT** create an issue with a green "New issue" button on top-right.

### Set a name

- Use `[Foo > Bar]` for an issue to write an article `Bar` in the group `Foo`. This issue should be a `Task`. When the article was already written, create a new issue `Plan` for it, hoisting the newly closed `Task` issue as its child.
- Use `[Foo > Bar] Add baz` to add info about `baz` to an article `Bar`. This issue should be a child to a plan `[Foo > Bar]`.
  - The first word must be a verb: add, improve, rewrite etc.
  - Do not duplicate the name of the article or things described in it here: it's already in `[]`.
- It's **not an arrow** `->`, it's just `>`.

### Set a type

- `Plan`: a group of sub-issues or sub-plans used to decompose a big task into a set of smaller ones; often times plans do not have an assignee because it's hard to pinpoint someone to them.
  - If it's a forever-issue for a certain article, put a `backlog` label on a plan, set no milestone, set no assignee.
[...]

PR #1415, opened 2025-11-23
chore(format): remove end-of-line spaces spotted by `remark-lint-no-trailing-spaces`

### LINKED ISSUE #1416: [Format] Do not use spaces at the end
Sometimes, when working with documentation and copying code snippets into examples, spaces may remain at the end of lines.

I suggest enabling the [remark-lint-no-trailing-spaces](https://github.com/Trott/remark-lint-no-trailing-spaces) plugin so that these issues no longer appear in the documentation. I looked at recent merger requests, and there, too, spaces at the end are removed.

Danil42Russia:
Closed #1416 

Unfortunately, there will still be gaps in some parts of the documentation due to the `.remarkignore` being ignored. If you don't mind, I would also remove the spaces at the end there.


anton-trunov:
@Danil42Russia thanks for you contribution

it's definitely not a feature (`feat`), please fix the PR title and open the corresponding issue as per the workflow: #1366

Danil42Russia:
@anton-trunov, fixed

package.json
```diff
@@ -30,6 +30,7 @@
     "remark-cli": "^12.0.1",
     "remark-frontmatter": "^5.0.0",
     "remark-gfm": "^4.0.1",
+    "remark-lint-no-trailing-spaces": "^4.0.3",
```
novusnota:
```suggestion
    "remark-lint-no-trailing-spaces": "=4.0.3",
```

Let's pin an exact version of this community-maintained package, just in case

novusnota (changes requested):
Thanks! The end-of-line spaces are handled by editorignore-compatible editors and plugins, but it's good to explicitly make the formatter do so too.

Regarding ignored files — don't :) The whitepapers there are frequently changed as of now, and their formatting is probably the last thing we'll do after that process ends.

package.json
Danil42Russia: fixed!

novusnota (dismissed):
I've formatted touched files once again. Weirdly enough, `remark-lint-no-trailing-spaces` does not remove the trailing spaces, only warns about them. You might consider removing it from package.json and reverting the package-lock.json changes.

In any case, we'll probably switch to a `.mjs` config for remark soon, to fine tune it and allow reuse in `scripts/`. There, I'll make it trim unnecessary spaces :)

Danil42Russia:
Yes, unfortunately, `remark-lint-no-trailing-spaces` cannot fix errors :(
But I suggest leaving it so that no new errors appear. Then it can be removed later.

verytactical:
Trying to gauge whether this should be merged or closed.

It's common to leave some spaces at the end of the line, and it'd be quite unfortunate if this has to be fixed by hand in every place, or else merge of a PR would be blocked.

Are there any cases when spaces at the end of the line hurt, i.e. might make some code not work or article to render incorrectly?

novusnota:
> Are there any cases when spaces at the end of the line hurt, i.e. might make some code not work or article to render incorrectly?

Never. Otherwise they'd be handled in the formatter, and they are not.

Currently, having an `.editorconfig` enables end-of-line space trimming in all editors that support it (incl. exotic ones like Helix). And a bit later, I'd like to introduce such trimming as part of the multi-step formatting pipeline when working on #1482.

Overall, I'm OK with merging this PR as long as we don't bring a new dependency in, since it'll only add noise in the formatter's output :/

Danil42Russia:
These spaces are unpleasant to see in the final documentation:

non-existent example:
Screenshot 2025-11-29 at 03 12 25

real example:
Screenshot 2025-11-29 at 03 34 43

I don't see any problem with fixing a few comments manually if it improves the user experience.

verytactical:
Sorry, I will be closing this, because spending the time on manual fixes is too expensive. I edited the issue to tell that we need this to be autofixable to be viable, and reassigned the issue to @novusnota.

PR #1425, opened 2025-11-24
feat(ci): unconditionally hide prior formatting PR comments, and only post a new one in case of failure

### LINKED ISSUE #1016: [Format] Wrong command recommended for formatting
Formatter recommends using `npx remark`, but it doesn't _have_ to be the same version of remark. It might be a global remark too.

### LINKED ISSUE #1284: [Format] If the formatting check was successful, it should hide all prior comments from the bot and mark those as resolved
https://github.com/int128/hide-comment-action/tree/main/src/queries

Or through a `action/github-script@v7` to fine tune it to our needs.

novusnota:
Closes #1284.
Also closes #1016, because it now recommends `npm run fmt:some` with passing files to it.

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -46,36 +46,53 @@ jobs:
           files: |
             **.md
             **.mdx
+          separator: ":"
 
       - name: Check formatting of MDX and Markdown files
         id: check-fmt
         env:
           ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
-        run: |
-          set -euo pipefail
-          if [ -z $(echo -e "${ALL_CHANGED_FILES[@]}" | tr -d '[:space:]') ]; then
-            echo -e '\nNo such files affected!'
-            exit 0
-          fi
-          echo -e '\nChecking formatting of the following MDX and Markdown files affected by this PR:\n'
-          for file in ${ALL_CHANGED_FILES}; do
-            echo "- $file"
-          done
-          FILES="$(echo -e "${ALL_CHANGED_FILES[@]}" | tr '\n' ' ')"
-          ( # try
-            npm run check:fmt:some -- $FILES >/dev/null 2>&1
-          ) || ( # catch
-            ERR="$(echo $?)"
-            # Comment right in the actions output
-            echo -e "\n\033[31mError:\033[0m Some files are not properly formatted!"
-            echo -e "1. Install necessary dependencies: \033[31mnpm ci\033[0m"
-            echo -e "2. Run this command to fix the issues: \033[31mnpm run fmt:some -- ${FILES}\033[0m"
-            # Prepare a comment on the PR
-            COMMENT="To fix the **formatting** issues:\n\n1. Install necessary dependencies: \`npm ci\`\n2. Then, run this command:\n\`\`\`shell\nnpm run fmt:some -- ${FILES}\n\`\`\`"
-            echo "COMMENT='$COMMENT'" >> "$GITHUB_ENV" # notice the ''
-            # Rethrow the exit code of the failed formatting check
-            exit $ERR
-          )
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        with:
+          script: |
+            const files = (process.env.ALL_CHANGED_FILES ?? '').trim().split(':').filter(Boolean);
+            if (files.length === 0) {
+              console.log('\nNo such files affected!');
+              process.exit(0);
+            }
+            console.log('\nChecking formatting of the following MDX and Markdown files affected by this PR:\n');
+            for (const file of files) {
+              console.log(`- ${file}`);
+            }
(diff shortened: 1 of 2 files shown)


.github/workflows/linter.yml
```diff
@@ -10,13 +10,13 @@ on:
...
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}-linter
   cancel-in-progress: true
 
 permissions:
   contents: read
   pull-requests: write
-  issues: write
+  # issues: write
```
novusnota, line 19: Commented out since it didn't seem necessary, will remove later if nothing fails

.github/workflows/linter.yml
```diff
@@ -60,27 +61,86 @@ jobs:
...
-          curl -s -o /dev/null -L -X POST \
-            -H "Accept: application/vnd.github+json" \
-            -H "Authorization: Bearer ${{ secrets.GITHUB_TOKEN }}" \
-            -H "X-GitHub-Api-Version: 2022-11-28" \
-            -d "$BODY" \
-            https://api.github.com/repos/${{ github.repository }}/issues/${{ github.event.number }}/comments
-          # Comment right in the actions output
-          echo -e "\nInstall necessary dependencies: \033[31mnpm ci\033[0m"
-          echo -e "Then, run this to fix formatting: \033[31mnpx remark -o --silent --silently-ignore ${FILES}\033[0m"
+          ( # try
```
novusnota: 1. Notes in the actions output are now rendered in the same step with the formatting check
2. All raw `npx` formatting commands got replaced with relevant "scripts" in `package.json`: both non-modifying checks and in-place formatting invocations

.github/workflows/linter.yml
```diff
@@ -60,27 +61,86 @@ jobs:
...
+            echo -e "1. Install necessary dependencies: \033[31mnpm ci\033[0m"
+            echo -e "2. Run this command to fix the issues: \033[31mnpm run fmt:some -- ${FILES}\033[0m"
+            # Prepare a comment on the PR
+            COMMENT="To fix the **formatting** issues:\n\n1. Install necessary dependencies: \`npm ci\`\n2. Then, run this command:\n\`\`\`shell\nnpm run fmt:some -- ${FILES}\n\`\`\`"
+            echo "COMMENT='$COMMENT'" >> "$GITHUB_ENV" # notice the ''
+            # Rethrow the exit code of the failed formatting check
+            exit $ERR
+          )
+
+      - name: Hide prior PR comments and issue a new one in case of failure
```
novusnota, line 97: If the check was successful, it hides all non-hidden/non-minimized comments from the bot that start with the same prefix.

Otherwise, it still hides all non-hidden comments, but for a different `minimizedReason`. Then it posts a new comment, which may use a slightly different formatting command if there are new or different `.mdx` files.

Note that GitHub's UI shows all reasons as "minimized", which is a known issue reported on Jul 2, 2022: https://github.com/orgs/community/discussions/19865.

.github/workflows/linter.yml
```diff
@@ -60,27 +61,86 @@ jobs:
...
+        if: ${{ !cancelled() && (github.event_name == 'pull_request' || github.event_name == 'pull_request_target') }}
+        env:
+          COMMENT: ${{ env.COMMENT }}
+          SUCCESS: ${{ steps.check-fmt.conclusion == 'failure' && 'false' || 'true' }}
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        with:
+          script: |
+            const success = JSON.parse(process.env.SUCCESS ?? 'false');
```
verytactical, line 106: Should be in `/scripts/`

novusnota: Or in `.github/scripts/` if they are meant only for GitHub CI, not for local use.

novusnota: Done!

.github/workflows/linter.yml
```diff
@@ -60,27 +61,32 @@ jobs:
...
+
+      - name: Hide prior PR comments and issue a new one in case of failure
+        if: ${{ !cancelled() && github.event_name == 'pull_request' && github.event_name != 'pull_request_target' }}
+        env:
+          COMMENT: ${{ env.COMMENT }}
+          SUCCESS: ${{ steps.check-fmt.conclusion == 'failure' && 'false' || 'true' }}
+        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7.1.0
+        with:
+          script: |
+            const { default: script } = await import('${{ github.workspace }}/.github/scripts/manage-pr-comments-after-check.mjs');
```
novusnota: Can't go much smaller than this

Re: https://github.com/actions/github-script/tree/v7.1.0#run-a-separate-file

novusnota (commented):
Let's convert all Bash in `linter.yml` to JavaScript while we're at it

UPD: or not. *Thinking*...

verytactical (approved):

PR #1443, opened 2025-11-26
feat: add bcp visuals

### LINKED ISSUE #1287: Visualize BCP and Catchain protocol interactions
We need to make interactive component that will demonstrate principles of how consensus works

Kaladin13:
Closes #1287


docs.json
```diff
@@ -524,6 +524,12 @@
               "foundations/actions/change-library"
             ]
           },
+          {
+            "group": "Consensus",
```
verytactical, line 528: (Mostly not to forget until the time we look at it with Anton)

Creating this group without any other content seems strange: the user would open the group, and would find a tool with no explanation, except for the reference to a very technical whitepaper.

Probably we should merge it without a change to `docs.json`.

Gusarich:
sorry but I'll use this pr to test if pitaya still works fine after a bunch of refactors 🤭🤭

PR #1504, opened 2025-12-01
chore(ci): decrease the amount of scripting in `bouncer.yml` and stop posting messages on new pushes

### LINKED ISSUE #1497: [Bouncer] Separate scripts
Yaml must not have embedded scripts.

novusnota:
Closes #1497.

Stopped posting messages on new pushes in the PR — they are in the workflow outputs now. This significantly reduced the JS code used in the workflow, to the point where it's not really a win even if we manage to extract it elsewhere.

Also made the wording of workflow outputs be a bit smoother ­— no more "until then your PR is doomed and you're failure" type stuff. It's now more chill and explanatory, but still concise.

### DIFF
--- a/.github/workflows/bouncer.yml
+++ b/.github/workflows/bouncer.yml
@@ -127,7 +127,6 @@ jobs:
 
   enforce-better-descriptions:
     name: "Title and description"
-    if: github.event.action == 'opened' || github.event.action == 'edited'
     runs-on: ubuntu-latest
     steps:
       # pr title check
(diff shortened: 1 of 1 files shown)


.github/workflows/bouncer.yml
```diff
@@ -5,11 +5,9 @@ env:
   # additions only
   MAX_ADDITIONS_FORKS: 500
   # on rare occasions maintainers need to edit a lot of things at once
-  MAX_ADDITIONS_DIRECT_BRANCHES: 800
+  MAX_ADDITIONS_DIRECT_BRANCHES: 1000
```
verytactical, line 8: Why do we keep raising this limit? We discussed internally this should be 500 both for external and internal PRs.

novusnota: Yeah, but if we don't make this CI mandatory (and we don't), there's a non-zero chance that folks would neglect other checks, thinking that they can merge stuff even when CI fails. I don't want that to happen, and for the time being, made a slightly larger leeway to see whether it's worth it. We can always roll back or hard cap everyone to 500 once the contribution and maintenance processes are more or less settled.

Oh, and we have a bunch (~60) of now-removed stub pages waiting to be written. Let's come back in, say, a month and change constraints if necessary :)

novusnota: Opened a task: https://github.com/ton-org/docs/issues/1521

PR #1514, opened 2025-12-02
fix(ci): formatter

### LINKED ISSUE #1512: Formatter command is broken
- adds two backquotes at the end
- adds a slash after non-folders `.mdx/`

https://github.com/ton-org/docs/pull/1510#issuecomment-3601330780

novusnota:
Closes #1512. Bash shenanigans strike again.

### DIFF
--- a/file with spaces.mdx
+++ b/file with spaces.mdx
@@ -1,5 +1,0 @@
----
-title: "Hey, I'm a file with spaces!"
----
-
-Too much spaces...


.github/workflows/linter.yml
```diff
@@ -109,11 +109,11 @@ jobs:
               console.log('There was a formatting error, but no comment was given, skipping...');
               process.exit(0);
             }
-            const prefix = rawCommentText.slice(1, 30);
+            const prefix = rawCommentText.slice(0, 30);
             await hidePriorCommentsWithPrefix({ github, context, exec, prefix, resolved: success });
             // Create a new comment in case of a new failure
             if (!success) {
-              const body = rawCommentText.slice(1, -1).replace(/\\n/g, '\n');
```
novusnota, line 116: Previously, that was done to circumvent bash shenanigans, which is no longer necessary after the move to JS

novusnota:
Ok, here's an issue:
1. `env` is always extra-escaped
2. `outputs` are not escaped, but if they are multi-line, they cannot be read by JavaScript

Possible solutions:
1. Resort to the `env` case, and do not try to process files in a fancy way. "KISS".
2. Try not to pass the long message around and instead compose it on the fly when needed — after all, all steps after the `changed-files` one know the list of files changed :)

I'll dabble with the 2nd one for a short bit, and if that fails, let's revert fancy changes and carefully sidestep Bash string escaping (as it was done before)

UPD: The 2nd approach works (?)
UPD2: Only works partially, but it might be unsafe, so let's not.

PR #1520, opened 2025-12-03
fix(whitepapers): redirects for PDFs

### LINKED ISSUE #1518: [Whitepapers] Fix redirects for pdfs
For instance, this does not work https://docs.ton.org/ton.pdf, but https://old-docs.ton.org/ton.pdf does. The same applies for the rest of the pdfs.

novusnota:
Closes #1518. We had similar redirects around 2-3 months ago, before their overhaul, but they didn't work then.

So, PDF redirects work locally, but not in the staging preview. The workaround could (probably) be to make symbolic links from the root. UPD: Nope, symbolic links do not fix this.

In theory, we can do the inverse: place real PDFs in the root and symbolically link to them from `/resources/pdfs/...`. But that might break other places which expect files in the `resources` directory. Let's just report that to Mintlify if we didn't do it yet.

P.S.: As a last resort, we could simply copy-paste those 5 PDFs. Not ideal, but Git should be able to recognize and deduplicate them. Right? UPD: Right, hashes from `git hash-object` are the same, which means that Git stores duplicate PDFs once, and both paths reference the same blob object. Shall we do that @anton-trunov @verytactical?

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -3203,6 +3203,11 @@
       "permanent": true
     },
     {
+      "source": "/ton.pdf",
+      "destination": "/resources/pdfs/ton.pdf",
+      "permanent": true
+    },
+    {
       "source": "/tblkch.pdf",
       "destination": "/resources/pdfs/tblkch.pdf",
       "permanent": true
--- a/docs.json
+++ b/docs.json
@@ -3203,11 +3203,6 @@
       "permanent": true
     },
     {
-      "source": "/ton.pdf",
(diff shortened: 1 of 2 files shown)


novusnota:
1. Let's see how fumadocs will fare here
2. In any case, we can just place PDFs under `public/`, and then reference from there instead of refering to relative-ish links to resource folders

Henceforth, this PR will be closed to be remade later on.

Issue #1577, opened 2025-12-10
[Tolk > Migrating from FunC  > Tolk-vs-TL-B] Remove article from documentation


aigerimu:
Remove the article from documentation because: 

1. all TL-B content is already covered in the TL-B section, and the same applies to Tolk.
2. it doesn’t fit any documentation type. It’s not a how-to, tutorial, ref, or explanation.

tolk-vm:
This is absolutely wrong.
If you ask me (the author of all Tolk documentation), "what's the most valuable article in all Tolk docs?" — I'll definitely mention this one.

I was asked many times questions like "How to provide a TL-B schema for a Tolk struct", or "Why can't we use TL-B to express ABI of a contract", that I've created a dedicated page revealing all the explanation. Now I refer such questions to this page.

To emphasize the essence of this page, I've even mentioned it in the announcement several days ago: https://t.me/tolk_lang/26

And about your points from the description:

\> 1. all TL-B content is already covered in the TL-B section, and the same applies to Tolk.
This is wrong: it does not cover TL-B, it compares TL-B and Tolk and explains all questions arised; "the same" does not apply anywhere. If you think it is, you misunderstood the article. Read it again.

\> 2. it doesn’t fit any documentation type. It’s not a how-to, tutorial, ref, or explanation.
This is wrong: it's the explanation.

PR #1578, opened 2025-12-10
chore: tolk-vs-tlb

aigerimu:
closes https://github.com/ton-org/docs/issues/1577


tolk-vm:
Absolutely wrong. See a comment in #1577

aigerimu:
@tolk-vm, since this page read more like a blog post than a reference page, its contents were moved to https://ton.org/en/tolk-vs-tl-b and were expanded upon there.

Hence, users will be redirected to that post instead of this page.

Issue #1583, opened 2025-12-11
[Tolk > Language features > Contract getters] Improve article in agreement with styleguide


aigerimu:
Put [article](https://docs.ton.org/languages/tolk/features/contract-getters) in agreement with styleguide

PR #1613, opened 2025-12-11
feat: tolk-contract-getters

aigerimu:
closes https://github.com/ton-org/docs/issues/1583

### DIFF
--- a/languages/tolk/features/contract-getters.mdx
+++ b/languages/tolk/features/contract-getters.mdx
@@ -72,7 +72,7 @@ get fun get_wallet_address(ownerAddress: address): address {
 
 ## Stack-based execution
 
-Get methods accept parameters and return values through the TVM stack because they are [off-chain](/from-ethereum#on-chain-get-methods).
+Any function in TVM takes its arguments from the stack and pushes return values onto it; get methods differ only in that they can be invoked [off-chain](/from-ethereum#on-chain-get-methods). When called off-chain, a get method does not persist state changes.
 
 A getter can return `int`, which is not serializable unlike `intN`. Returning a structure pushes each field onto the stack as a separate value. Client libraries such as [Blueprint](/contract-dev/blueprint/overview) parse get method responses using a tuple reader.
 


languages/tolk/features/contract-getters.mdx
```diff
@@ -63,51 +56,24 @@ get fun get_wallet_data(): JettonWalletDataReply {
...
-
-## Get methods operate via the stack, not serialization
-
-Get methods accept parameters and return values via the TVM stack, since they are called off-chain.
+## Stack-based execution
 
-This is why **a getter can return `int`** (although it's not serializable, unlike `intN`, see [numbers](/languages/tolk/types/numbers)).
-Returning a structure pushes its fields onto the stack as N separate values.
-Client libraries such as Blueprint parse get‑method responses using a tuple reader.
+Get methods accept parameters and return values through the TVM stack because they are [off-chain](/from-ethereum#on-chain-get-methods).
```
novusnota:
```suggestion
```

Any function takes arguments from and pushes return values onto the TVM stack. It's just that getters, apart from being callable on-chain from other functions of the same contract, can be queried off-chain.

It's probably better to say that when called off-chain, getters will not affect the state of the contract, but this seems to be beyond the descriptions here.

I'd consider adding some content from https://docs.tact-lang.org/book/functions/#get to the bottom, since Tolk does take some inspiration from Tact's getter implementation. There's a `@method_id` attribute in Tolk that allows custom method IDs — it seems like to use it, the `get` keyword has to be omitted, but that should be double-checked: https://github.com/ton-blockchain/ton/blob/f7f0c90831b0df495a8dc9ea7413d3a6c0e74830/tolk-tester/tests/methods-tests.tolk#L74-L78

aigerimu: resolved as: 

Any function in TVM takes its arguments from the stack and pushes return values onto it; get methods differ only in that they can be invoked [off-chain](/from-ethereum#on-chain-get-methods). When called off-chain, a get method does not persist state changes.

Issue #1621, opened 2025-12-12
[Tolk > Syntax details > Pattern matching]


aigerimu:
Put [article](https://docs.ton.org/languages/tolk/syntax/pattern-matching) in agreement with styleguide

PR #1630, opened 2025-12-13
feat: tolk-pattern-matching

aigerimu:
closes https://github.com/ton-org/docs/issues/1621

### DIFF
--- a/languages/tolk/syntax/pattern-matching.mdx
+++ b/languages/tolk/syntax/pattern-matching.mdx
@@ -162,5 +162,3 @@ match (someColor) {
     else => {}
 }
 ```
-
-


languages/tolk/syntax/pattern-matching.mdx
```diff
@@ -122,8 +121,9 @@ val nextValue = match (curValue) {
 
 Rules:
 
-- only constant expressions may appear before `=>`
-- `else` is required for match expressions and optional for statements
+- only constant expressions are allowed before `=>`
+- `else` is required when `match` is used as an expression
+- `else` is optional when match is used as a statement
```
novusnota:
```suggestion
- `else` is optional when `match` is used as a statement
```

languages/tolk/syntax/pattern-matching.mdx
```diff
@@ -140,13 +140,11 @@ val nextValue = match (curValue) {
 }
 ```
 
-<Aside type="tip">
-  All comparable types supported — for instance, addresses and enums.
-</Aside>
+All comparable types are supported, including addresses and enums
```
novusnota:
```suggestion
All comparable types are supported, including addresses and enums.
```

languages/tolk/syntax/pattern-matching.mdx
```diff
@@ -140,13 +140,11 @@ val nextValue = match (curValue) {
...
 
 ## `match` for enums
 
-Pattern matching on [enums](/languages/tolk/types/enums) requires coverage of all cases (_exhaustive_):
+Pattern matching on [enums](/languages/tolk/types/enums) requires coverage of all cases, exhaustive:
```
novusnota:
```suggestion
Pattern matching on [enums](/languages/tolk/types/enums) requires exhaustive coverage of all cases:
```

languages/tolk/syntax/pattern-matching.mdx
```diff
@@ -165,9 +163,4 @@ match (someColor) {
 }
 ```
 
-<Aside
-  type="note"
-  title={"Familiar with Rust's enum and match?"}
->
-  Note that `match` is similar in spirit, but `enum` values in Tolk are backed by integers; union types are more general and expressive.
-</Aside>
+Compared to Rust, `match` is similar in spirit, but `enum` values in Tolk are backed by integers, while union types are more general and expressive.
```
novusnota: This should've stayed as a side comment, since Rust is external to this documentation and, compared to TypeScript, is less familiar to a regular developer.

However, from the prior info on this page, Rust users can see that Tolk's `enum` and `match` are inferior to what they have — thus, I think this comment is redundant and should be removed.

Issue #1576, opened 2025-12-09
[Tolk > Language features > Sending messages] Improve article in agreement with styleguide


aigerimu:
Put [article](https://docs.ton.org/languages/tolk/features/message-sending) in agreement with styleguide

PR #1662, opened 2025-12-17
feat: tolk-msgs

aigerimu:
closes https://github.com/ton-org/docs/issues/1576

### DIFF
--- a/languages/tolk/features/standard-library.mdx
+++ b/languages/tolk/features/standard-library.mdx
@@ -1,34 +1,31 @@
 ---
-title: "Standard library of Tolk"
+title: "Standard library"
 sidebarTitle: "Standard library"
 ---
 
 import { Aside } from '/snippets/aside.jsx';
 
-Tolk provides many functions for everyday use, available out of the box.
-They are called "standard library", or "stdlib".
+Tolk provides a standard library or stdlib that includes commonly used functions.
 
 <Aside type="tip">
-  An IDE auto-completes all functions.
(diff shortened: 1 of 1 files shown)


languages/tolk/features/message-sending.mdx
```diff
@@ -94,11 +65,11 @@ struct CreateMessageOptions<TBody> {
...
 ```
 
-**That's indeed the TypeScript way** — but it works at compile-time.
+Each option represents a valid way to specify the message destination. The selected form is resolved at compile time.
 
-## Deployment, stateInit, and workchains
+## Deployment and `StateInit`
 
-Consider the following example. A contract "jetton minter" deploys a "jetton wallet". The wallet's code and initial data are known:
+Consider a contract that deploys another contract. For example, a jetton minter deploys a jetton wallet. The wallet code and its initial data are known:
```
novusnota:
```suggestion
Consider a contract that deploys another contract. For example, a jetton minter deploying a jetton wallet. The wallet code and its initial data are known:
```

languages/tolk/features/message-sending.mdx
```diff
@@ -234,11 +198,11 @@ createMessage({
 });
 ```
 
-Therefore, do not pass `body: obj.toCell()`, pass just `body: obj`, let the compiler take care of everything.
+Therefore, do not pass `body: obj.toCell()`. Pass `body: obj`, and the compiler chooses the correct encoding.
```
novusnota:
```suggestion
Therefore, do not pass `body: obj.toCell()`. Pass `body: obj`, and the compiler will choose the optimal and correct encoding.
```

PR #1054, opened 2025-10-30
feat: add toncenter overview and rate limit pages

laviniat1996:
Closes #1044
Closes #1041

Added new overview.mdx page for TON Center, covering hosted access and API environments
Added new rate-limit.mdx page explaining plan tiers, default limits, and rate-limit behavior.

PR #1825, opened 2026-02-10
fix(api): adjust scripts, regenerate toncenter api pages, correct recent changes

### LINKED ISSUE #1797: [Ecosystem > APIs > TON Center] Refactor pages order and naming for API Center refference docs
- During development, we faced fragmented page names and positions. Once all pages are added(according to the [plan](https://github.com/ton-org/docs/issues/1020)), refactor the content and build a precise structure with simple naming conventions.

novusnota:
Towards #1054 — reverts some changes made there, regenerates pages for API v3, etc.
Closes #1797

### DIFF
--- a/.github/workflows/generate-api-tables.yml
+++ b/.github/workflows/generate-api-tables.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install pyyaml
+          pip install pyyaml==6.0.3
           corepack enable
           npm ci
 
--- a/docs.json
+++ b/docs.json
@@ -31,6 +31,7 @@
       "copy",
       "chatgpt",
       "claude",
+      "grok",
       "perplexity",
       "mcp",
       "cursor",


.github/workflows/generate-api-tables.yml
```diff
@@ -14,23 +20,32 @@ jobs:
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version: ${{ env.NODE_VERSION }}
+          cache: "npm"
+
       - name: Install dependencies
-        run: pip install pyyaml
-      
+        run: |
+          pip install pyyaml
```
Danil42Russia: it is better to explicitly lock the dependency version :)

.github/workflows/generate-api-tables.yml
```diff
@@ -14,23 +20,32 @@ jobs:
...
       - uses: actions/checkout@v4
-      
+
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.11'
-      
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - uses: actions/setup-node@v4
```
Danil42Russia, line 28: it seems that we can also upgrade the action version to `v6` :)
Well, others can too...

.github/workflows/generate-api-tables.yml
novusnota: Good catch, thanks!

.github/workflows/generate-api-tables.yml
novusnota: Perhaps so, but in v5 they change default Node version from 20 to 24. Sure, the action setup tells it to use version 20 explicitly, but I remember having unexpected version switches in the past, espectially on their macOS runners, so let's keep things as is.

aigerimu (approved):

PR #1842, opened 2026-02-12
Update contract-getters.mdx

bodasooqa:
Just a couple of lines above, it says that camelCase should be used for getter names.

### DIFF
--- a/languages/tolk/features/contract-getters.mdx
+++ b/languages/tolk/features/contract-getters.mdx
@@ -44,7 +44,7 @@ struct JettonWalletDataReply {
     jettonWalletCode: cell
 }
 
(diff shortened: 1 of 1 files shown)


bodasooqa:
image

novusnota:
> Just a couple of lines above, it says that camelCase should be used for getter names.

Yes, but immediately after the paragraph says that this convention may be overridden when matching standard TEPs (TON Enhancement Protocols). In particular, jetton wallets need the snake-cased way of writing getters because that affects the resulting method_id and it's what the ecosystem expects from jetton contracts.

Sorry, but I will be closing this PR. Thank you for finding the time to create it :)

PR #1846, opened 2026-02-16
Fix front matter formatting in start-here.mdx

AnonymousScript007:


novusnota:
Sorry, but this PR breaks the frontmatter. The [frontmatter extension](https://github.com/remarkjs/remark-frontmatter?tab=readme-ov-file#example-different-markers-and-fences) for Markdown (CommonMark) and MDX denotes triple-dash markers as YAML frontmatter block, whereas your commit breaks --- markers by insering an # before them.

Thus, I'll close this PR. Thanks for trying though!

PR #1868, opened 2026-02-19
feat(api): add pagination page for v2 api

### LINKED ISSUE #1024: [TON Center > API/v2 > Pagination]
Create a page that provides a clear explanation of Pagination in the TON Center API/v2. Explain how the developer of the TON Center expect their pagination to be used.

References: 
- https://docs.stripe.com/api/pagination
- https://www.twilio.com/docs/usage/twilios-response#pagination
- https://docs.slack.dev/apis/web-api/pagination/

laviniat1996:
closes #1024 

This PR adds a new page which contains in-depth details about how API v2 pagination works.

## Summary by CodeRabbit

* **Documentation**
  * Added a Toncenter v2 pagination guide covering cursor-based pagination for transaction endpoints, parameter requirements, inclusive/exclusive cursor behavior, termination signals, URL-encoding notes, example requests/responses, and JavaScript pagination scripts.
  * Added the new pagination page to the TON Center → API v2 documentation navigation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/ton-org/docs/pull/1868?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


laviniat1996:
@reveloper So I added parameters here as well for each endpoint, the params are already listed in the autogenerated api pages section. Should we keep them here as well or not? I'm not sure

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,670 @@
...
+## getTransactions
+
+Returns the transaction history of a given account. Results are ordered by logical time in descending order (newest first).
+
+Pagination uses `lt` and `hash` as a cursor pair. Both must be provided together. These values come from the last transaction in the previous response.
+
+### Parameters
+
+| Parameter  | Type    | Required | Description                                                                    |
+| ---------- | ------- | :------: | ------------------------------------------------------------------------------ |
```
reveloper: Let's add `No` for each optional parameter in the table.

Image

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,670 @@
...
+This returns transactions between the two logical times, which is useful when syncing only recent activity.
+
+## getBlockTransactions
+
+Returns a list of short transaction identifiers for all transactions in a given block.
+
+Pagination uses `after_lt` and `after_hash` as a cursor pair. Both must be provided together. The response includes an `incomplete` field that indicates whether more transactions remain in the block.
+
+### Parameters
+
```
reveloper, line 267: Set `No` for optional requirements:

Image

reveloper (changes requested):

reveloper:
@laviniat1996, nice question!
Regarding request params, I think we may provide a limited description in a table, i.e. for 
[getTransactions](https://companyname-a7d5b98e-v2-pagination.mintlify.app/ecosystem/api/toncenter/v2-pagination#gettransactions) should be a table with `lt` and `hash`, which is directly related to the pagination context. 
Instead of listing all parameters, we may set a link or a caution with "See reference docs for entire parameter list."

This approach would be helpful to keep an up-to-date description and field list, in case they change over time.

laviniat1996:
it fails a check because  this PR  has over 600 new lines, but it should be okay because there are a lot of scripts in there that take up space

laviniat1996:
@novusnota could you please review this one as well when you have time? it has one failed check, because the PR is more than 600 lines long, but it's this long because there are a lot of code examples

delovoyhomie (commented):
Nice page overall — really glad to see inclusive vs. exclusive cursor called out explicitly. A few inline notes: the main one is a termination bug in the first full pagination script that silently stops after page 2 for any account with more than 9 transactions, plus a couple of consistency nits between the two block-transactions sections.

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,685 @@
...
+        }
+      }
+
+      if (lt && hash) {
+        result.shift();
+      }
+
+      allTransactions.push(...result);
+
+      if (result.length < 5) break;
```
delovoyhomie: Could we move the `if (result.length < 5) break;` check above `result.shift()`? On every page after the first, a full API response of 5 items is shifted down to 4, so the condition is always true and the loop exits after page 2. The sample output of 9 transactions happens to hide this because `5 + (5 − 1) = 9` lands exactly at the cutoff, but on any account with more than 9 transactions the script silently stops early. Something like `const LIMIT = 5;` used for both the `limit` query param and the termination check would also keep the two from drifting later.

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,685 @@
+---
+title: "Pagination"
+sidebarTitle: "Pagination"
+---
+
+import { Aside } from '/snippets/aside.jsx';
+
+The v2 API uses cursor-based pagination to retrieve large result sets in batches. Instead of specifying a page number, the identifier of the last record received is passed to the API, which returns the next batch starting from that point. This ensures consistent results even when new data is being added.
```
delovoyhomie: Could we tighten the second sentence? `getTransactions` uses an inclusive cursor, so callers actually do see the boundary transaction twice and have to dedupe — that is exactly what the first pagination script does with `result.shift()` a few sections down, and it is called out in the `Inclusive cursor` Aside. As written the intro reads as if cursor-based pagination guarantees duplicate-free pages, which then contradicts the rest of the page. Something like "Because the cursor points at an existing record instead of a page number, new data inserted at the head of the list does not shift already-seen pages" would match the actual behaviour.

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,685 @@
...
+        ]
+      }
+    }
+    ```
+
+    The `incomplete` field is `false`. All transactions in this block have been retrieved.
+
+    <Aside
+      type="note"
+      title="Exclusive cursor"
```
delovoyhomie: For `getBlockTransactions` the `Exclusive cursor` Aside sits after `</Steps>`, but here it is inside Step 3 of the `getBlockTransactionsExt` walk-through. Could we keep the placement the same in both sections? I would lean toward after `</Steps>` in both cases, since it is meta-guidance about the cursor rather than part of the three-step walk-through itself.

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,685 @@
...
+
+  Total transactions in block: 4
+  ```
+</Accordion>
+
+## getBlockTransactionsExt
+
+Returns full transaction details for all transactions in a given block. Pagination works the same way as [`getBlockTransactions`](#getblocktransactions), using `after_lt` and `after_hash` as cursors with the `incomplete` field as the end of results signal. The only difference is that this endpoint returns complete transaction objects instead of short identifiers.
+
+### Pagination parameters
```
delovoyhomie, line 478: The `Pagination parameters` table here is byte-identical to the one under `## getBlockTransactions`. Could we replace it with a one-liner, e.g. `Pagination parameters are identical to [\`getBlockTransactions\`](#getblocktransactions).`? Right now any future change to `after_lt` / `after_hash` / `count` semantics has to be made in both places, and the two copies can silently drift.

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,685 @@
...
+      title="Inclusive cursor"
+    >
+      The transaction passed as `lt`/`hash` is included as the first result in the next response. In the responses above, transaction `66784070000010` appears as the last item on page 1 and again as the first item on page 2. When implementing pagination, skip the first result on subsequent pages to avoid processing the same transaction twice.
+    </Aside>
+  </Step>
+
+  <Step
+    title="Repeat until the last page"
+  >
+    Continue extracting the cursor from the last transaction and fetching the next page. When the response returns fewer transactions than the `limit`, all results have been retrieved.
```
delovoyhomie: Could we spell out the two end-of-data cases for the inclusive cursor here? Because `lt`/`hash` is inclusive, on a subsequent page a full page actually delivers `limit − 1` new items, and a response that contains only the cursor transaction (1 item) is also an end signal. Without that bridge, `Response returns fewer results than limit` in the top table reads the same for the first page and for subsequent pages, which is exactly what the bug in the full pagination script above trips on.

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,677 @@
+---
+title: "Pagination"
+sidebarTitle: "Pagination"
```
delovoyhomie, line 3: I'd suggest the following:

```mdx
---
title: "How to paginate Toncenter v2 results"
sidebarTitle: "Pagination"
---
```

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,677 @@
...
+All other v2 endpoints return single objects or fixed results and do not support pagination.
+
+<Aside
+  type="caution"
+  title="URL encoding"
+>
+  Transaction hashes are base64-encoded and may contain `+`, `/`, and `=` characters. These must be URL-encoded when passed as query parameters: `+` → `%2B`, `/` → `%2F`, `=` → `%3D`. Most HTTP libraries handle this automatically, but this is important when constructing URLs manually.
+</Aside>
+
+## getTransactions
```
delovoyhomie: `## getTransactions`, `## getBlockTransactions` (line 256), and `## getBlockTransactionsExt` (line 472) put a code identifier directly into the heading without code font. Headings normally must not contain styling, but the style guide explicitly allows code font for identifiers on reference-style entries (see [Formatting restrictions](https://docs.ton.org/contribute/style-guide-extended#74-formatting-restrictions)):

```md
## `getTransactions`
```

that also makes the slug/anchor (`#gettransactions`) read as an identifier in cross-page links.

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,677 @@
...
+| `limit`   | integer | Maximum number of transactions to return.                                      |
+
+<Aside
+  type="note"
+  title="Full parameter list"
+>
+  This table only lists pagination-related parameters. See the [`getTransactions` reference](/ecosystem/api/toncenter/v2/accounts/list-account-transactions) for all available parameters.
+</Aside>
+
+### Paginating through transactions
```
delovoyhomie: `### Paginating through transactions` (and the matching H3s at lines 279 and 480) use gerund form. The style guide asks headings to use the imperative, it means that  gerunds shouldn't appear on H2/H3 https://docs.ton.org/contribute/style-guide-extended#72-gerunds-and-labels

```md
### Paginate through transactions
### Paginate through block transactions
### Paginate through extended block transactions
```

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,677 @@
...
+
+<Steps>
+  <Step
+    title="Fetch the first page"
+  >
+    Send a request with only `address` and `limit`. No cursor parameters are needed for the first page.
+
+    ```bash
+    curl "https://toncenter.com/api/v2/getTransactions?address=EQDtFpEwcFAEcRe5mLVh2N6C0x-_hJEM7W61_JLnSF74p4q2&limit=5"
```
delovoyhomie, line 60: examples should default to testnet  `https://toncenter.com/…` hits mainnet. The style guide says examples SHOULD run end-to-end on testnet by default https://docs.ton.org/contribute/style-guide-extended#101-general-rules

but I am not into it very much, so probably mainnet is appropriate here

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,677 @@
...
+
+### Fetching a specific range
+
+Use `to_lt` to stop pagination at a specific logical time instead of iterating through the entire history:
+
+```bash
+curl "https://toncenter.com/api/v2/getTransactions?address=EQDtFpEwcFAEcRe5mLVh2N6C0x-_hJEM7W61_JLnSF74p4q2&limit=20&lt=66822814000007&to_lt=66784070000010"
+```
+
+This returns transactions between the two logical times, which is useful when syncing only recent activity.
```
delovoyhomie: `which is useful when syncing only recent activity` — `useful` is in the banned vague-adjective list (see [Voice and tone](https://docs.ton.org/contribute/style-guide-extended#4-voice-and-tone))

I'd suggest

```suggestion
This returns only transactions between the two logical times, so a sync job can re-fetch a bounded window rather than walking the full history.
```

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,677 @@
...
+    All results have been retrieved when the response returns fewer transactions than the `limit`. Because the cursor is inclusive, a full subsequent page delivers `limit - 1` new transactions (the first item is the cursor duplicate), so a response containing only the cursor transaction (1 item) also counts as an end signal.
+  </Step>
+</Steps>
+
+<Accordion
+  title="Full pagination script"
+>
+  ```javascript
+  const address = "EQDtFpEwcFAEcRe5mLVh2N6C0x-_hJEM7W61_JLnSF74p4q2";
+  const LIMIT = 5;
```
delovoyhomie: the reader is expected to change `address` and `LIMIT`, but the script also has `EQDtFpEwcFAEcRe5mLVh2N6C0x-_hJEM7W61_JLnSF74p4q2` hard-coded. The style guide says runnable examples must collect adjustable values in one clearly visible block near the top, using `<ANGLE_CASE>` placeholders in `UPPER_SNAKE` (or top-level constants) inside language code (see [Code and command examples](https://docs.ton.org/contribute/style-guide-extended#101-general-rules)

```suggestion
  // Replace with a testnet account address you want to paginate.
  const ADDRESS = "<TESTNET_ACCOUNT_ADDRESS>";
  const LIMIT = 5;
  const API_BASE = "https://testnet.toncenter.com/api/v2";
```

then reference `ADDRESS`, `LIMIT`, `API_BASE` in the loop body and define `<TESTNET_ACCOUNT_ADDRESS>` once below the block, same shape applies to the two later scripts

ecosystem/api/toncenter/v2-pagination.mdx
```diff
@@ -0,0 +1,677 @@
...
+| [`getBlockTransactions`](/ecosystem/api/toncenter/v2/blocks/list-block-transactions)                     | `after_lt` + `after_hash` | `incomplete` field is `false`               |
+| [`getBlockTransactionsExt`](/ecosystem/api/toncenter/v2/blocks/list-block-transactions-extended-details) | `after_lt` + `after_hash` | `incomplete` field is `false`               |
+
+All other v2 endpoints return single objects or fixed results and do not support pagination.
+
+<Aside
+  type="caution"
+  title="URL encoding"
+>
+  Transaction hashes are base64-encoded and may contain `+`, `/`, and `=` characters. These must be URL-encoded when passed as query parameters: `+` → `%2B`, `/` → `%2F`, `=` → `%3D`. Most HTTP libraries handle this automatically, but this is important when constructing URLs manually.
```
delovoyhomie: `but this is important when constructing URLs manually` - `important` here is a vague intensifier without specifics. the style guide asks to replace hedges/intensifiers with the concrete consequence (see [Tautology, pleonasm, throat-clearing](https://docs.ton.org/contribute/style-guide-extended#59-avoid-tautology-pleonasm-throat-clearing-and-circular-references))

```suggestion
  Transaction hashes are base64-encoded and may contain `+`, `/`, and `=` characters. These must be URL-encoded when passed as query parameters: `+` → `%2B`, `/` → `%2F`, `=` → `%3D`. Most HTTP libraries handle this automatically. When constructing URLs manually, encode `+`, `/`, and `=` explicitly or the cursor will not match the server-side value.
```

reveloper:
Hello! Thank you for the review!
Since Lavinia no more work with us, I will handle this personally, possible during 1-2 weeks.

ecosystem/api/toncenter/v2-pagination.mdx
reveloper: Since the mainnet API is a bit more stable and consistency is guaranteed more, I think such an approach will work better in numerous cases.
However, if we have all examples in the testnet, I may change this for consistency. I'm just not sure we should follow this for API services.

@novusnota, can you share your opinion?

ecosystem/api/toncenter/v2-pagination.mdx
reveloper: I think it is more optimal here to keep the example address in the comment, to speed up the example for readers who haven't received the address yet.
I've kept the mainnet address since it works better.

reveloper:
@delovoyhomie updated.

novusnota:
@coderabbitai please, run a deep technical review of the new page against the V2 API sources:

- https://github.com/toncenter/ton-http-api — Python version
- https://github.com/toncenter/ton-http-api-cpp — C++ version

PR #1903, opened 2026-03-05
feat(appkit): how to swap

### LINKED ISSUE #1857: [Ecosystem > TON Connect > AppKit > How to swap assets]
with Omniston

### LINKED ISSUE #1986: [Ecosystem > AppKit] DeDust swap provider

novusnota:
Closes #1857 (How-to for swaps with AppKit)
Closes #1986 (DeDust swap provider for AppKit)

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive AppKit token swap documentation page with React and TypeScript examples for obtaining quotes and building swap transactions, including slippage and deadline configuration
  * Introduced DeDust as a newly supported swap provider
  * Clarified swap provider selection behavior with defaults and provider targeting
  * Updated Toncoin documentation with refined balance check guidance


novusnota:
@coderabbitai hi

novusnota:
@coderabbitai summary

ecosystem/appkit/swap.mdx
```diff
@@ -0,0 +1,382 @@
...
+      quote,
+      userAddress: selectedWallet?.getAddress() ?? '<TON_WALLET_ADDRESS>',
+      slippageBps: 100, // 1%
+      deadline: Math.floor(Date.now() / 1000) + 600, // 10 minutes
+    });
+
+    // Step 3: Sign and send via TON Connect
+    const result = await sendTransaction(kit, transaction);
+    console.log('Swap transaction sent:', result);
+  }
```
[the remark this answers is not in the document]
novusnota: Hmm, not sure that's needed here... Like you wrote, there's a "Funds at risk" caution already.

ecosystem/appkit/swap.mdx
```diff
@@ -0,0 +1,382 @@
...
+      amount,
+      network: Network.mainnet(),
+    });
+    const { mutateAsync: buildTransaction, isPending: isBuilding } =
+      useBuildSwapTransaction();
+    const { mutateAsync: sendTransaction, isPending: isSending } =
+      useSendTransaction();
+
+    const handleSwap = async () => {
+      if (!quote || !wallet) return;
```
delovoyhomie: This snippet imports `useAddress` and stores its result as `address`, but the guard on this line reads `wallet` - which is never declared or imported anywhere in the example. As written, the first click on the Swap button throws `ReferenceError: wallet is not defined` before any swap logic runs. Either rename to `address` to match the rest of `handleSwap`, or pull in `useWallet` and use it consistently.

ecosystem/appkit/swap.mdx
```diff
@@ -0,0 +1,382 @@
...
+
+      // Build the swap transaction from the quote
+      const transaction = await buildTransaction({
+        quote,
+        userAddress: address ?? '<TON_WALLET_ADDRESS>',
```
delovoyhomie: `userAddress: address ?? '<TON_WALLET_ADDRESS>'` will quietly pass the literal string `<TON_WALLET_ADDRESS>` into `buildTransaction` when no wallet is connected. Best case the SDK throws a confusing parse error deep inside the provider; worst case the provider builds something nobody asked for. The guard above already short-circuits when there's no wallet, so the fallback is also dead code - dropping the `??` (and fixing the guard per the comment above) makes the contract cleaner: if you got here, `address` is real. The TypeScript example uses the same placeholder pattern on line 277 and should be updated together.

ecosystem/appkit/swap.mdx
```diff
@@ -0,0 +1,382 @@
...
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify({
+        quote: params.quote,
+        userAddress: params.userAddress,
+        destinationAddress: params.destinationAddress,
+        slippageBps: params.slippageBps,
+        deadline: params.deadline,
+      }),
+    });
+    return response.json();
```
delovoyhomie, line 354: `return response.json()` only does the right thing if `https://api.example-dex.com/build` happens to respond in the exact `TransactionRequest` shape AppKit expects (messages array with `address`, `amount`, `payload`, `validUntil`, etc.). No real DEX API ships that shape, so a reader who copies this template gets a provider that compiles, registers cleanly, and then dies the first time anyone actually calls `buildSwapTransaction`. Worth either a short comment about mapping the DEX response into `TransactionRequest`, or an inline sketch of the target shape.

ecosystem/appkit/swap.mdx
```diff
@@ -0,0 +1,382 @@
...
+
+<CodeGroup>
+  ```tsx title="React" icon="react"
+  import {
+    Network,
+    useSwapQuote,
+    useBuildSwapTransaction,
+    useSendTransaction,
+    useAddress,
+  } from '@ton/appkit-react';
```
delovoyhomie, line 193: The quote example a few sections up imports `Network` from `@ton/appkit`, but this snippet pulls it from `@ton/appkit-react` along with the hooks. If both packages re-export the same symbol, that's worth saying explicitly so a reader bouncing between sections doesn't have to guess; if only one of them does, one of the two examples won't compile. Worth picking a single source per framework and using it across all snippets.

ecosystem/appkit/init.mdx
```diff
@@ -180,32 +180,51 @@ For local development and testing, use the [manifest file from this demo dApp](h
...
-    import { AppKit, Network } from '@ton/appkit';
+    import { AppKit } from '@ton/appkit';
 
-    // Omniston as a swap provider
+    // Swap providers
     import { OmnistonSwapProvider } from '@ton/appkit/swap/omniston';
+    import { DeDustSwapProvider } from '@ton/appkit/swap/dedust';
+
+    // Tonstakers as a staking provider
+    import { createTonstakersProvider } from '@ton/appkit/staking/tonstakers';
```
delovoyhomie, line 219: `createTonstakersProvider` is imported but never appears in the `providers: [...]` array below, so the import sits dead in the example. Either drop it (since this section is about swap providers), or actually include `createTonstakersProvider({...})` in the array - the step is titled "Set up one or more DeFi providers", which would justify showing staking alongside the swap providers if that was the intent.

ecosystem/appkit/swap.mdx
TrueCarry: @novusnota bump

ecosystem/appkit/init.mdx
TrueCarry: @novusnota bump

novusnota:
@coderabbitai review

PR #1975, opened 2026-03-19
feat(api): update v3 API field descriptions

### LINKED ISSUE #1974: [TON Center > API/v3 > full field description]
Update all fields for indexer (v3 API) in order for each field to be properly defined and easy to understand by devs, similar to how v2 fields were updated.

laviniat1996:
closes #1974


novusnota:
Major thanks for annotating the schema! That said, changes diverge from the upstream TON Center API v3 specification. A proper solution requires careful changes to the upstream code, followed by a port of the resulting specification back to the docs.

I'll close this PR, but will preserve the branch so some parts can be reused later. Thanks again!

Issue #232, opened 2025-09-16
[Foundations > Blockchain sharding]


verytactical:
- add a disclaimer: not the same thing as contract sharding
- https://docs.ton.org/v3/documentation/smart-contracts/shards/shards-intro
- https://docs.ton.org/v3/documentation/smart-contracts/shards/infinity-sharding-paradigm

Issue #675, opened 2025-10-11
[Links] Introduce a better link checker and replace current `link-rot` CI


novusnota:
What we cannot do with the current one:
1. Selectively check links in the files given in the PR
2. Ignore certain routes or paths. For example, any of the existing API route pages are not present on disk, marking them as "broken" in the existing checker. The solution is to create such pages manually, see: #660
3. Check relative heading anchors (`#header-id`)

Only the 3rd issue poses some problems in the long run, but they are _currently_ negligible as documentation moves quite fast.

Consider implementing this with `remark` and:
- https://github.com/remarkjs/remark-validate-links
- (optional) https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-no-undefined-references

novusnota:
This is more cumbersome than I've originally thought, so I'll postpone that a little (to "Future work") and be quickly resolving other things. I think that the present link check will be enough for the time being.

Notes to self: try having multiple configurations.

novusnota:
We need this now to fix links in whitepapers and, possibly, find other invalid relative (anchor) links — hence the return of `"prio: high"`

PR #1979, opened 2026-03-19
fix: broken link

coalus:
just a little fix, current url leads to 404 page

Towards #232


novusnota (approved):
Thanks for spotting it! We'll try to improve our link checker as per #675 somewhat sooner :)

PR #1982, opened 2026-03-20
feat(tma): add "launch parameters" and "init data" pages

### LINKED ISSUE #1981: [Ecosystem > TMA] Create pages for launch parameters & init data explanation in TMA section
rn docs have no pages explaining what data Telegram passes to a Mini App or how to authenticate users.

This should be added:
- New "Launch parameters" page: what Telegram passes at startup, parameter reference table, start parameter for deep linking
- New "Init data" page: what init data is, how to send it to the server, validation, field reference

Follow-up: Next PRs will cover UI components, dev workflow, and SDK overview.

coalus:
resolves #1981

### DIFF
--- a/ecosystem/tma/init-data.mdx
+++ b/ecosystem/tma/init-data.mdx
@@ -1,0 +1,252 @@
+---
+
+title: "Init data"
+
+---
+
+Init data is a signed payload that the Telegram client generates when launching a Mini App. It is transmitted in the [`tgWebAppData`](/ecosystem/tma/launch-parameters#tgwebappdata) launch parameter and contains information about the current user, chat context, and a cryptographic signature.
+
+Telegram signs the payload with the bot token, so a server can verify the signature and trust the content without additional authentication.
+
+## Retrieving init data
+
+Extract init data from launch parameters using the [@tma.js/sdk](https://www.npmjs.com/package/@tma.js/sdk):
+
+```typescript
+import { retrieveLaunchParams } from '@tma.js/sdk';
+
+const { tgWebAppData: initData } = retrieveLaunchParams();
+```
+
+## Sending init data to the server
+
+Pass raw init data in the `Authorization` header with every request:
+
+```typescript
+import { retrieveRawInitData } from '@tma.js/sdk';
+
+const initDataRaw = retrieveRawInitData();
+
+fetch('https://example.com/api', {
+  method: 'POST',
+  headers: {
+    Authorization: `tma ${initDataRaw}`,
+  },
+});
+```
+
+The header format is `tma <INIT_DATA_RAW>`, where `<INIT_DATA_RAW>` is the raw query string from the `tgWebAppData` launch parameter.
+
+## Validating init data
+
+Validation confirms that Telegram issued the payload and that no one tampered with it. Two methods are available.
+
+### Using the bot token
+
+This method requires the bot token that is bound to the Mini App.
+
+1. Parse the init data as query parameters. Separate the `hash` value and remove it from the set.
+1. Sort the remaining key-value pairs alphabetically and join them as `key=value` strings separated by `\n`.
+1. Create an HMAC-SHA256 of the bot token using the string `WebAppData` as the key.
+1. Create an HMAC-SHA256 of the sorted pairs string (from step 2) using the result of step 3 as the key. Convert the output to a hex string.
+1. Compare the result with the `hash` value from step 1. If they match, the init data is valid.
+
+<Aside
+  type="caution"
+>
+  The HMAC-SHA256 result from step 3 is a raw byte array. Do not convert it to a hex string before using it as the key in step 4.
+</Aside>
+
+### Using the Telegram public key
+
+This method validates init data without the bot token — only the bot identifier is needed. It uses Ed25519 signature verification.
+
+Telegram provides two public keys:
+
+| Environment | Ed25519 public key (hex)                                           |
+| ----------- | ------------------------------------------------------------------ |
+| Production  | `e7bf03a2fa4602af4580703d88dda5bb59f32ed8b02a56c187fe7d34caed242d` |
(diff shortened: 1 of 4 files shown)


ecosystem/tma/init-data.mdx
```diff
@@ -0,0 +1,254 @@
...
+};
+
+const app = express();
+app.use(authMiddleware);
+
+app.get('/', (_req, res) => {
+  res.json(res.locals.initData);
+});
+
+app.use(((err, _req, res) => {
```
delovoyhomie, line 129: Express only treats a middleware as an error handler when its arity is 4, and this one takes three params. So `next(new Error('Unauthorized'))` from `authMiddleware` above won't land here - it'll fall through to Express's default handler (HTML page in prod, stack trace in dev) and the `res.status(500).json(...)` never fires. Adding a fourth parameter (`_next: NextFunction`) fixes the arity. Separately, returning 500 for a missing or malformed `Authorization` header feels off - that's a 401. Worth branching so legitimate server bugs and client auth failures don't share a status code.

ecosystem/tma/launch-parameters.mdx
```diff
@@ -0,0 +1,56 @@
...
+https://t.me/<BOT_USERNAME>?startattach=<VALUE>
+https://t.me/<BOT_USERNAME>/<APP_NAME>?startapp=<VALUE>
+```
+
+In both cases, `tgWebAppStartParam` contains `<VALUE>`. This value is also duplicated in the [`start_param`](/ecosystem/tma/init-data#init-data-parameters) field of init data.
+
+### Restrictions
+
+- Allowed characters: `A-Z`, `a-z`, `0-9`, `_`, `-`
+- Maximum length: 512 characters
```
delovoyhomie, line 52: Is 512 actually the Telegram-side cap on `startapp` / `startattach`? My recollection is 64 characters (same ceiling as the classic `start_parameter` in the Bot API). If this was raised recently a link to the source would help; otherwise the docs and the client will disagree the first time someone passes a 300-char payload and the launch silently drops it.

ecosystem/tma/init-data.mdx
```diff
@@ -0,0 +1,254 @@
...
+Telegram provides two public keys:
+
+| Environment | Ed25519 public key (hex)                                           |
+| ----------- | ------------------------------------------------------------------ |
+| Production  | `e7bf03a2fa4602af4580703d88dda5bb59f32ed8b02a56c187fe7d34caed242d` |
+| Test        | `40055058a4ee38156a06562e52eece92a771bcd8346a8c4615cb7376eddf72ec` |
+
+1. Parse the init data as query parameters. Separate the `signature` value (base64-encoded) and remove both `hash` and `signature` from the set.
+1. Sort the remaining key-value pairs alphabetically and join them as `key=value` strings separated by `\n`.
+1. Prepend the string `<BOT_ID>:WebAppData\n` to the result from step 2.
```
delovoyhomie, line 75: The selling point of this method is "no bot token needed, just the bot identifier" - but the page never says where `<BOT_ID>` comes from. It's the integer before the colon in the bot token, and also the `id` returned by `getMe`. One sentence making that explicit here would close the loop for readers who landed on this section precisely because they don't have (or don't want to expose) the token.

ecosystem/tma/launch-parameters.mdx
```diff
@@ -0,0 +1,56 @@
...
+| `tgWebAppPlatform`     | `string`  | Platform identifier: `android`, `ios`, `macos`, `tdesktop`, `weba`, or `web`.                                                                                                                                                     |
+| `tgWebAppThemeParams`  | `string`  | Telegram client theme (JSON-encoded). Contains color keys such as `bg_color`, `text_color`, `button_color`, `header_bg_color`, and others. See [Theming](https://docs.telegram-mini-apps.com/platform/theming) for the full list. |
+| `tgWebAppStartParam`   | `string`  | Custom value from the `startattach` or `startapp` query parameter in the launch link. See [Start parameter](#start-parameter).                                                                                                    |
+| `tgWebAppShowSettings` | `boolean` | Internal flag for the settings button. Not relevant for Mini App developers.                                                                                                                                                      |
```
delovoyhomie, line 34: If the flag is "internal" and "not relevant for Mini App developers", what's it doing in the public reference? Either drop the row, or rewrite so it explains its own presence - e.g. "the Telegram client writes it but only the Telegram-owned settings UI consumes it, so you can safely ignore it". As-is the entry tells readers to ignore something without saying why they're reading about it.

ecosystem/tma/launch-parameters.mdx
```diff
@@ -0,0 +1,56 @@
...
+| `tgWebAppBotInline`    | `boolean` | Present when the Mini App is launched in inline mode.                                                                                                                                                                             |
```
delovoyhomie, line 35: `tgWebAppBotInline` and `tgWebAppFullscreen` are listed as `boolean`, but what the Telegram client actually writes into the URL hash is either the literal string (`"1"` / `"true"`) or the key is absent. `retrieveLaunchParams` coerces, so callers going through the SDK see real booleans - fine. A reader parsing the hash themselves (also a documented path, via the raw accessors) will be typechecking against the wrong shape. Either note that the Type column reflects post-SDK values, or split wire vs. parsed types.

novusnota:
Thank you for your work ❤️

Unfortunately, embedding generic TMA documentation into TON documentation fell outside the future scope and plans. If anything, there might be a need for TON-related tutorial(s) that use TMA docs when building TMAs, plus brief links to those TMA docs from application-oriented overview pages. Exact implementation will depend on the future shape and structure of the docs, which might include a dedicated section for dApp developers.

In any case, although this PR will be closed, your work is recognized and kindly appreciated. Besides, your non-TMA pull requests are not forgotten — they will be reviewed and merged promptly during the following week or two. Please, stay in touch with the team at that time. Thanks!

PR #1990, opened 2026-03-22
feat(tma): add theming, viewport, and UI components pages

### LINKED ISSUE #1989: [Ecosystem > TMA] Add theming, viewport, and UI components pages
This should be added:

- New "Theming" page: how to retrieve and apply Telegram theme colors, color reference
- New "Viewport" page: viewport properties, fullscreen mode, closing and swipe behavior
- New "UI components" page: native controls provided by Telegram (Back Button, Main Button, Settings Button, Popup)

Follow-up: Next PRs will cover dev workflow and SDK overview

coalus:
resolves #1989

### DIFF
--- a/ecosystem/tma/theming.mdx
+++ b/ecosystem/tma/theming.mdx
@@ -1,0 +1,63 @@
+---
+title: "Theming"
+---
+
+Mini Apps are designed to look native inside Telegram. To achieve this, the Telegram client provides a set of theme colors that the app can apply to its UI.
+
+## Retrieving theme colors
+
+Theme colors are available in two ways.
+
+**Launch parameter.** The [`tgWebAppThemeParams`](/ecosystem/tma/launch-parameters) launch parameter contains a JSON-encoded object with color values. Parse it at startup to apply the initial theme.
+
+**Runtime updates.** The Telegram client can change the theme while the app is running (for example, when the user switches between light and dark mode). Call the [`web_app_request_theme`](https://docs.telegram-mini-apps.com/platform/methods#web-app-request-theme) method and listen for the [`theme_changed`](https://docs.telegram-mini-apps.com/platform/events#theme-changed) event to receive the updated colors.
+
+## Color reference
+
+The theme object contains the following optional properties:
+
+| Property                    | Description                                      |
+| --------------------------- | ------------------------------------------------ |
+| `bg_color`                  | Main background color.                           |
+| `secondary_bg_color`        | Secondary background color.                      |
+| `text_color`                | Primary text color.                              |
+| `hint_color`                | Hint and placeholder text color.                 |
+| `link_color`                | Link color.                                      |
+| `button_color`              | Default button background color.                 |
+| `button_text_color`         | Default button text color.                       |
+| `accent_text_color`         | Accent text color (highlights, active elements). |
+| `destructive_text_color`    | Destructive action text color (delete, remove).  |
+| `header_bg_color`           | Header background color.                         |
+| `bottom_bar_bg_color`       | Bottom bar background color.                     |
+| `section_bg_color`          | Section background color.                        |
+| `section_header_text_color` | Section header text color.                       |
+| `section_separator_color`   | Section separator line color.                    |
+| `subtitle_text_color`       | Subtitle text color.                             |
+
+Example value:
+
+```json
+{
+  "accent_text_color": "#6ab2f2",
+  "bg_color": "#17212b",
+  "button_color": "#5288c1",
+  "button_text_color": "#ffffff",
+  "bottom_bar_bg_color": "#ffffff",
+  "destructive_text_color": "#ec3942",
+  "header_bg_color": "#17212b",
+  "hint_color": "#708499",
+  "link_color": "#6ab3f3",
+  "secondary_bg_color": "#232e3c",
+  "section_bg_color": "#17212b",
+  "section_header_text_color": "#6ab3f3",
+  "subtitle_text_color": "#708499",
+  "text_color": "#f5f5f5"
+}
+```
+
+## Background and header colors
+
+A Mini App is displayed inside a native container that has a header bar and a body area. Telegram provides methods to change the colors of both:
+
+- [`web_app_set_header_color`](https://docs.telegram-mini-apps.com/platform/methods#web-app-set-header-color) — set the header color using a theme key or a custom RGB string.
+- [`web_app_set_background_color`](https://docs.telegram-mini-apps.com/platform/methods#web-app-set-background-color) — set the body background color.
--- a/ecosystem/tma/ui-components.mdx
+++ b/ecosystem/tma/ui-components.mdx
@@ -2,6 +2,8 @@
 title: "UI components"
 ---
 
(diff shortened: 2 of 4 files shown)


ecosystem/tma/ui-components.mdx
```diff
@@ -0,0 +1,65 @@
+---
+title: "UI components"
+---
+
+Telegram provides a set of native UI components that Mini Apps can show and configure. These components are rendered by the Telegram client outside the WebView, so they match the platform look and feel.
```
novusnota, line 7:
```suggestion

import { Aside } from '/snippets/aside.jsx';

Telegram provides a set of native UI components that Mini Apps can show and configure. These components are rendered by the Telegram client outside the WebView, so they match the platform look and feel.
```

Usage of `Aside` requires an import — this is not a built-in component of Mintlify. By the way, we have docs on our in-house, custom components: https://docs.ton.org/contribute/snippets/overview#custom-components

ecosystem/tma/viewport.mdx
```diff
@@ -0,0 +1,48 @@
+---
+title: "Viewport"
+---
+
```
novusnota, line 4: Aside import is missing, see: https://github.com/ton-org/docs/pull/1990/changes#r2971623702

coalus:
Linting check should be ok after #1982 is merged

ecosystem/tma/viewport.mdx
```diff
@@ -0,0 +1,50 @@
...
+## Viewport properties
+
+| Property     | Type      | Description                                                         |
+| ------------ | --------- | ------------------------------------------------------------------- |
+| `width`      | `number`  | Width of the visible area in pixels.                                |
+| `height`     | `number`  | Height of the visible area in pixels.                               |
+| `stability`  | `boolean` | `true` when the viewport size is not expected to change imminently. |
+| `expansion`  | `boolean` | `true` when the Mini App has reached its maximum height.            |
+| `fullscreen` | `boolean` | `true` when the Mini App is displayed in fullscreen mode.           |
```
delovoyhomie, line 17: Two things trip up in this table. First, the property names don't match what the raw `viewport_changed` event actually delivers - the payload carries `height`, `width`, `is_state_stable`, and `is_expanded`. A reader who wires up the listener linked below and tries to read `payload.stability` or `payload.expansion` will get `undefined`. If the names here are the tma.js SDK's camelCased / simplified accessors, that's worth calling out; if the table is describing the Telegram-side shape, the keys need to change. Second, `fullscreen` isn't part of `viewport_changed` at all - it ships in its own `fullscreen_changed` event (correctly linked further down on line 38). Bundling it into a table anchored to `viewport_changed` quietly conflates two separate event sources.

ecosystem/tma/theming.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+Example value:
+
+```json
+{
+  "accent_text_color": "#6ab2f2",
+  "bg_color": "#17212b",
+  "button_color": "#5288c1",
+  "button_text_color": "#ffffff",
+  "bottom_bar_bg_color": "#ffffff",
```
delovoyhomie, line 45: `"bottom_bar_bg_color": "#ffffff"` in the middle of what is otherwise a dark-theme example looks like a paste slip - everything else here (`#17212b`, `#232e3c`, `#708499`, `#f5f5f5`) is straight out of Telegram's dark palette. Copied as-is, the bottom bar will flash solid white against the dark body in any app that uses this snippet verbatim. Probably wants the corresponding dark value (Telegram's own dark theme serves `#17212b` for this key).

ecosystem/tma/theming.mdx
```diff
@@ -0,0 +1,63 @@
...
+  "header_bg_color": "#17212b",
+  "hint_color": "#708499",
+  "link_color": "#6ab3f3",
+  "secondary_bg_color": "#232e3c",
+  "section_bg_color": "#17212b",
+  "section_header_text_color": "#6ab3f3",
+  "subtitle_text_color": "#708499",
+  "text_color": "#f5f5f5"
+}
+```
```
delovoyhomie, line 56: For a page titled "Theming", this covers half the journey - how to *retrieve* the theme but never how to *apply* it. The conventional pattern (and what every Mini Apps SDK does under the hood) is to map each key to a CSS variable - `--tg-theme-bg-color`, `--tg-theme-text-color`, and so on - so the rest of the stylesheet can reference `var(--tg-theme-bg-color)` and update automatically on `theme_changed`. A short snippet here (either a raw `document.documentElement.style.setProperty` loop or the tma.js `bindThemeParamsCssVars()` equivalent) would take this page from "reference" to "actually usable end-to-end".

ecosystem/tma/ui-components.mdx
```diff
@@ -0,0 +1,67 @@
...
+
+<Aside type="tip">
+  If the action triggered by the Main Button takes time to complete, show the built-in loader. This signals to the user that the app is processing the request.
+</Aside>
+
+## Settings Button
+
+The Settings Button appears in the top-right menu of the Mini App. It has no built-in behavior — the Mini App decides what happens when the user taps it.
+
+Available since Mini Apps API version `6.10`.
```
delovoyhomie, line 37: Singling out the Settings Button with a version note while the other four components get nothing reads as an oversight rather than a deliberate call. Back Button / Main Button have shipped since 6.1, Popup since 6.2, Haptic feedback since 6.1, and there are newer ones that didn't make this page (Secondary Button, Bottom Bar). If minimum API version matters for readers targeting older clients, add it consistently across every component; if it doesn't, drop the line and let the external reference handle version history.

ecosystem/tma/theming.mdx
```diff
@@ -0,0 +1,63 @@
...
+
+## Background and header colors
+
+A Mini App is displayed inside a native container that has a header bar and a body area. Telegram provides methods to change the colors of both:
+
+- [`web_app_set_header_color`](https://docs.telegram-mini-apps.com/platform/methods#web-app-set-header-color) — set the header color using a theme key or a custom RGB string.
+- [`web_app_set_background_color`](https://docs.telegram-mini-apps.com/platform/methods#web-app-set-background-color) — set the body background color.
```
delovoyhomie, line 63: The theme table above lists `bottom_bar_bg_color`, but this "Background and header colors" section only covers the header and body setters - the matching `web_app_set_bottom_bar_color` method is missing. A reader who sees the key and wants to recolor the bottom bar to match the rest of their app won't find the link here. Either add a third bullet for it, or rename the section to scope it explicitly to header/background and leave bottom-bar coloring out of range.

novusnota:
Thank you for your work ❤️

Unfortunately, embedding generic TMA documentation into TON documentation fell outside the future scope and plans. If anything, there might be a need for TON-related tutorial(s) that use TMA docs when building TMAs, plus brief links to those TMA docs from application-oriented overview pages. Exact implementation will depend on the future shape and structure of the docs, which might include a dedicated section for dApp developers.

In any case, although this PR will be closed, your work is recognized and kindly appreciated. Besides, your non-TMA pull requests are not forgotten — they will be reviewed and merged promptly during the following week or two. Please, stay in touch with the team at that time. Thanks!

PR #2013, opened 2026-03-24
fix(internal-infra): highlighting of custom languages in code blocks

### LINKED ISSUE #2014: Fix highlighting for custom languages: Tolk, TASM, TL-B, etc.
A follow-up on #1301. Basically, highlighting works locally and in staging environments, but seems to be disabled in production. Investigating this with Mintlify team.

novusnota:
Closes #2014.

Thing is, highlighting works locally and in staging, but does not appear in production for some reason. Might be a bug in Mintlify. Resolving...

### DIFF
--- a/docs.json
+++ b/docs.json
@@ -325,6 +325,7 @@
             "group": "Staking",
             "pages": [
               "ecosystem/staking/overview",
+              "ecosystem/staking/stake-calculation",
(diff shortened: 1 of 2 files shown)


novusnota:
Problem got resolved on the Mintlify's side, so experiments here are no longer needed.

PR #2016, opened 2026-03-24
fix(foundations): reserve can fail even with +16 flag set

### LINKED ISSUE #2002: [Foundations > Actions > Reserve] Incorrect reserve action failure description
# Reserve algorithm description issue

Documentation claims that on failure, reserve action will be skipped.

https://github.com/ton-org/docs/blob/45160fd71f10eb409ea81b4547e5dad19af264a3/foundations/actions/reserve.mdx?plain=1#L39-L43

However, this is not true, and reserve action fails just like any other.
It returns error code like any other action:https://github.com/ton-blockchain/ton/blob/a31025f39ed0dae5f6799280133624dc3a23cefb/crypto/block/transaction.cpp#L3105-L3114

And the error code handler is common for all the action types too.

In fact, there is no direct way to "skip" the failed reserve action like it is possible in message action.
Closest to skipping would be using `RESERVE_AT_MOST` flag, but in this case reserve action just can't fail, but not skipped.

novusnota:
Closes #2002. Also, adjusts the wording on the `/foundations/messages/internal#bounce` page a little.

## Summary by CodeRabbit

* **Documentation**
  * Clarified numeric formatting (uses literal 0) in reserve docs.
  * Refined reserve failure description to state that failed reservations explicitly enter the documented bounce phase.
  * Reworded error-handling language to consistently describe bounce-phase outcomes.
  * Updated bounce eligibility rules: compute-phase errors tied to uncommitted state, and action-phase rules now distinguish bounce behavior by action type rather than a single-mode condition.


foundations/messages/internal.mdx
```diff
@@ -93,8 +93,8 @@ A bounce message is used to inform the sender that handling of their message fai
 - there is enough Toncoin to send it left from handling the failed message,
 - and the message is bounceable,
 - and when either
-  - a message throws an error from the [compute phase](/foundations/phases#compute-phase) during message processing and state was not committed by [`COMMIT`](/tvm/instructions#f80f-commit) TVM instruction,
-  - or a message throws an error in the [action phase](/foundations/phases#action-phase) and mode [`SendBounceIfActionFail`](/foundations/messages/modes) is used.
+  - a contract throws an error from the [compute phase](/foundations/phases#compute-phase) during message processing, and the state was not committed by [`COMMIT`](/tvm/instructions#f80f-commit) TVM instruction,
+  - or an action fails in the [action phase](/foundations/phases#action-phase) with the `+16` flag enabled, for example [`SendBounceIfActionFail`](/foundations/messages/modes) in case of message sends or [`ReserveBounceIfActionFail`](/foundations/actions/reserve#modes) in case of reservations.
```
novusnota:
```suggestion
  - or an action fails in the [action phase](/foundations/phases#action-phase) with the `+16` flag enabled: [`SendBounceIfActionFail`](/foundations/messages/modes) in case of message sends or [`ReserveBounceIfActionFail`](/foundations/actions/reserve#modes) in case of reservations.
```

foundations/actions/reserve.mdx
```diff
@@ -31,16 +31,15 @@ Notation:
...
+  - if after the storage phase, account's balance is less than the value of incoming message with bounce flag set to `false`, then `original_balance` is set to 0;
   - otherwise, `original_balance` equals account's balance before the compute phase minus incoming message value.
 - `remaining_balance` – account's balance before the reservation action.
 - `reserve` – the final amount to be reserved.
 
 The algorithm is as follows:
 
 1. Check that `mode` has flag `ReserveBounceIfActionFail`:
-   - if so, then in case of any failure the action phase will be interrupted and the bounce phase will be initiated;
-   - if not, then in case of any failure the reservation action will be skipped.
```
aigerimu, line 43: why did you remove this: 
```
- if not, then in case of any failure the reservation action will be skipped.
```

aigerimu (dismissed):

aigerimu (approved):

delovoyhomie (approved):

foundations/actions/reserve.mdx
novusnota: Because it was not correct: the reservation action will not be skipped with or without the +16 flag (`ReserveBounceIfActionFail`), and the failure in such action will always lead to the failure of the action phase. It's just that _with_ the flag there'll be a bounce phase too.

That was the main point in the issue closed by this PR :)

PR #2052, opened 2026-04-03
feat(foundations): migrate ADNL documentation

### LINKED ISSUE #1035: [Foundations > Network] ADNL
Describe what it is, how it forks.

Describe what is a public and private overlay

coalus:
resolves #1035 

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive ADNL (Abstract Datagram Network Layer) overview covering addressing, key usage, handshake, encryption, packet integrity, and security notes
  * Published detailed ADNL over TCP and ADNL over UDP guides (handshake, packet formats, channel establishment, and message semantics)
  * Reorganized network protocols under a new "Blockchain foundations" group
  * Updated legacy ADNL redirects to the new foundation pages


docs.json
```diff
@@ -1385,27 +1393,27 @@
...
     {
       "source": "/v3/documentation/network/protocols/adnl/overview",
-      "destination": "https://old-docs.ton.org/v3/documentation/network/protocols/adnl/overview",
+      "destination": "/foundations/network/adnl",
       "permanent": true
     },
     {
       "source": "/v3/documentation/network/protocols/adnl/low-level",
-      "destination": "https://old-docs.ton.org/v3/documentation/network/protocols/adnl/low-level",
+      "destination": "/foundations/network/adnl",
```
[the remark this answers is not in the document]
coalus: The old low-level.mdx was mostly about peer identity/cryptosystems/TCP handshake - all merged into adnl.mdx. Redirect seems to be correct.

foundations/network/adnl-tcp.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+## Key ID calculation
+
+The key ID is the SHA-256 hash of the serialized TL schema. For ed25519 keys:
+
+```tl
+pub.ed25519 key:int256 = PublicKey -- ID c6b41348
+```
+
+The key ID is `SHA-256([0xC6, 0xB4, 0x13, 0x48] || public_key)` — a 36-byte input (4-byte TL prefix + 32-byte key).
```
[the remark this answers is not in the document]
novusnota: @coderabbitai Hmm, it might be given in little-endian order already. Please, triple-check.

foundations/network/adnl-udp.mdx
```diff
@@ -0,0 +1,189 @@
...
+}
+```
+
+Decode the ed25519 key from base64 and convert the IP integer to dotted-decimal format (for example, `1091897261` becomes `65.21.7.173`).
+
+## Channel creation
+
+To establish a communication channel, send two messages in the initial packet:
+
+1. **Create channel** (`adnl.message.createChannel`):
```
aigerimu: @coalus In the old doc, there is a TL serialization steps and examples for `adnl.message.createChannel` and `adnl.message.query`, but they are removed here. Was this intentional? If so, why?

foundations/network/adnl.mdx
```diff
@@ -0,0 +1,133 @@
...
+| `buffer`  | `length - 64` bytes     | Payload                                     |
+| `hash`    | 32 bytes                | `SHA-256(nonce \|\| buffer)` for integrity  |
+
+The entire structure is encrypted using the corresponding AES-CTR instance.
+
+The receiver decrypts the first 4 bytes to determine the length, then reads exactly that number of bytes and verifies the hash. If verification fails, the connection must be dropped.
+
+After a successful handshake, the server sends the first datagram to the client with an empty buffer.
+
+- Practical examples of [ADNL over TCP](/foundations/network/adnl-tcp), including liteserver communication, ping/pong, and smart contract method calls.
```
delovoyhomie: This bullet sits between the `After a successful handshake, the server sends the first datagram...` sentence (the last line of `### Datagram`) and the `### Security considerations` heading, with no surrounding list. It reads as an orphan `see also`. Could we either move it into the `## See also` block at the bottom of the page, or rewrite it as a short sentence — something like `For concrete examples (liteserver communication, ping/pong, smart contract method calls), see [ADNL over TCP](/foundations/network/adnl-tcp).` — so it does not look like a list item that lost its list?

foundations/network/adnl.mdx
```diff
@@ -0,0 +1,133 @@
...
+
+  The encryption key is derived from both the static `secret` and `SHA-256(aes_params)`. Since `aes_params` is random for each session, this results in a unique encryption key per connection. However, the concatenation-based derivation algorithm combines subarrays of the secret and the hash and is considered suboptimal by modern standards.
+
+- Datagram nonce
+
+  In CTR mode, AES operates as a stream cipher. If an attacker knows the plaintext, the bit-flipping attack becomes possible. The nonce mitigates this risk. Even if an attacker reconstructs the key stream and computes a valid `SHA-256(buffer)`, they cannot forge a valid hash without knowing the random nonce.
+
+## P2P protocol (ADNL over UDP)
+
+[In the UDP variant](/foundations/network/adnl-udp), communication begins simultaneously with sending initial data. If the initiator sends a _create channel_ message, a channel key is calculated, and the peer confirms creation.
```
delovoyhomie: This `## P2P protocol (ADNL over UDP)` section is two sentences, while `## Client-server protocol (ADNL over TCP)` above it has full `Handshake`, `Datagram`, and `Security considerations` subsections with byte-level tables. Could we either move the TCP deep-dive into [/foundations/network/adnl-tcp](/foundations/network/adnl-tcp) and keep `adnl.mdx` as a concise spec overview with pointers to both child pages, or add symmetric depth for UDP here (at least `Channel`, `Packet`, and a short security note)? As-is, the page spends roughly 80 lines on TCP specifics and 5 on UDP, which makes it read like `ADNL TCP with a UDP appendix` rather than a neutral protocol overview.

foundations/network/adnl-tcp.mdx
```diff
@@ -0,0 +1,231 @@
...
+| Cipher A | 0-31        | 64-79                         | Server encrypts, client decrypts |
+| Cipher B | 32-63       | 80-95                         | Client encrypts, server decrypts |
+
+The client sends a handshake packet:
+
+| Field                     | Size      |
+| ------------------------- | --------- |
+| Server key ID             | 32 bytes  |
+| Client ed25519 public key | 32 bytes  |
+| SHA-256 of the 160 bytes  | 32 bytes  |
```
delovoyhomie: the handshake field names here drift from the spec page at [/foundations/network/adnl#handshake](/foundations/network/adnl#handshake):

| This page | `adnl.mdx` |
| --- | --- |
| `Server key ID` | `receiver_address` |
| `Client ed25519 public key` | `sender_public` |
| `SHA-256 of the 160 bytes` | `SHA-256(aes_params)` |
| `Encrypted 160 bytes` | `E(aes_params)` |

wdyt does it make sense to align the two pages on one set of names? I would lean toward the `adnl.mdx` terms, since that is the normative spec page and also the form used in the `block.tlb` / reference implementation. rn a reader cross-referencing the two pages has to map four names each direction.

foundations/network/adnl-tcp.mdx
```diff
@@ -0,0 +1,231 @@
...
+
+[Implementation example](https://github.com/xssnick/tonutils-go/blob/2b5e5a0e6ceaf3f28309b0833cb45de81c580acc/liteclient/crypto.go#L16).
+
+## Handshake encryption
+
+The 160-byte session parameters are encrypted using an AES-CTR cipher derived from the SHA-256 hash of the 160 bytes and the [ECDH shared key](#ecdh-shared-key):
+
+```text
+key = shared_key[0:16] || hash[16:32]   // 16 bytes each
+iv  = hash[0:4] || shared_key[20:32]   // 4 + 12 bytes
```
delovoyhomie: Same terminology drift as the handshake table — the derivation here:

```
key = shared_key[0:16] || hash[16:32]
iv  = hash[0:4] || shared_key[20:32]
```

appears on [adnl.mdx:65–69](/foundations/network/adnl#handshake) as:

```
key   = secret[0..16] || hash[16..32]
nonce = hash[0..4]   || secret[20..32]
```

That is `shared_key`/`secret`, `iv`/`nonce`, and `[0:16]`/`[0..16]` — three separate divergences for literally the same three lines on two sibling pages. Could we pick one convention and mirror it on both pages?

foundations/network/adnl-tcp.mdx
```diff
@@ -0,0 +1,231 @@
...
+```
+
+[Implementation example](https://github.com/xssnick/tonutils-go/blob/2b5e5a0e6ceaf3f28309b0833cb45de81c580acc/liteclient/connection.go#L361).
+
+## ECDH shared key
+
+The shared key is computed from one party's private key and the other party's public key using Elliptic Curve Diffie-Hellman.
+
+Simplified example of the Diffie-Hellman principle (using small numbers):
```
delovoyhomie: Could we drop the simplified Diffie–Hellman walkthrough (`g=5`, `p=23`, `a=6`, ...) here, or move it into a collapsed `<Accordion>`? Per [`contribute/style-guide.mdx`](https://github.com/ton-org/docs/blob/main/contribute/style-guide.mdx) the default audience is `experienced software developers new to blockchain` — generic DH with toy numbers is well below that bar, and a Wikipedia link on `Elliptic Curve Diffie-Hellman` covers anyone who wants a refresher. The TON-specific part (`curve25519`, how ADNL actually uses the shared key bytes) is what this section should be answer-first about.

foundations/network/adnl-udp.mdx
```diff
@@ -0,0 +1,189 @@
...
+Key 2: shared_key (reversed byte order)
+```
+
+To determine which key encrypts outgoing messages and which decrypts incoming messages, compare the [key IDs](/foundations/network/adnl-tcp#key-id-calculation) of both parties' public keys as uint256:
+
+| Condition                        | Outgoing encryption | Incoming decryption |
+| -------------------------------- | ------------------- | ------------------- |
+| Server ID \< Client ID           | Key 1               | Key 2               |
+| Server ID > Client ID            | Key 2               | Key 1               |
+| Server ID = Client ID (unlikely) | Key 1               | Key 1               |
```
delovoyhomie: `Server ID = Client ID` is a peer talking to itself — that is not just `unlikely`, it is a degenerate case that should not occur between distinct peers. And the `Key 1 / Key 1` assignment means using the same key for both outgoing encryption and incoming decryption, which on its own is a bad configuration rather than a valid fallback. Could we either drop this row, or reword the condition to something like `Same peer ID (not applicable in practice)` so readers do not treat it as a supported branch of the comparison?

content/foundations/network/adnl.mdx
```diff
@@ -0,0 +1,133 @@
+---
+title: "Abstract Datagram Network Layer"
+sidebarTitle: "ADNL"
+---
+
+import { Aside } from "/snippets/aside.jsx";
+
+The Abstract Datagram Network Layer (ADNL) is the core networking protocol of the TON network. It is a peer-to-peer, unreliable datagram protocol that operates over both [UDP](https://en.wikipedia.org/wiki/User_Datagram_Protocol) and [TCP](https://en.wikipedia.org/wiki/Transmission_Control_Protocol).
+
+Higher-level protocols such as [RLDP](https://old-docs.ton.org/v3/documentation/network/protocols/rldp) and [DHT](https://old-docs.ton.org/v3/documentation/network/protocols/dht/overview) are built on top of ADNL.
```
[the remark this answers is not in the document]
aigerimu: let's change after merging RLDP/DHT PR, or we can do redirects? @novusnota @coalus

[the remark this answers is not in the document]
coalus: wdyt if we merge this PR first, then the RLDP/DHT PR, and after that I'll open a quick follow-up PR to swap the old-docs URLs for internal ones

foundations/network/adnl-udp.mdx
coalus: not intentional - trimmed them during migration just by mistake. restored both rn

content/foundations/network/adnl.mdx
```diff
@@ -0,0 +1,58 @@
...
```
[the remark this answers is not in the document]
coalus: intentional for the PR split - will swap to internal links in a follow-up PR after the RLDP/DHT PR lands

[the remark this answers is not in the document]
novusnota: Fair suggestion, but we've already decided to just have those links for a short while right until the follow-up PR, which will remove them.

foundations/network/adnl-tcp.mdx
```diff
@@ -0,0 +1,251 @@
...
+## Connection establishment
+
+Prerequisites:
+
+- Server IP, port, and public key from the [global config](https://ton-blockchain.github.io/global.config.json);
+- A freshly generated Ed25519 private and public key pair.
+
+<Aside type="tip">
+  The IP address in the config is a decimal integer. Convert it to dotted-decimal IPv4 format, for example using [this tool](https://www.browserling.com/tools/dec-to-ip). The public key is base64-encoded.
+</Aside>
```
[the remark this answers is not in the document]
novusnota: Fair, let's remove that link and instead just briefly tell what's needed to be done.

Besides, "The public key is base64-encoded" note is out of place, it should be added right next to the 2nd prerequisite that introduces "a freshly generated ed25519 private and public key pair".

novusnota (approved):
@coalus Thank you for the initial PR and follow-ups! ❤️

Sorry for the massive delay in its review process — there was a big move to a proper Fumadocs setup from Mintlify, apart from other things 🙏

I've applied a bunch of edits and will be merging this shortly.

PR #2079, opened 2026-04-10
feat(standard): migrate ton connect cookbook examples

### LINKED ISSUE #2078: [Standard contracts] Add TON Connect examples to jetton/NFT transfer pages
Migrate TON Connect cookbook examples into existing transfer pages

coalus:
resolves #2078 

## Summary by CodeRabbit

* **Documentation**
  * Added guides for Jetton transfers and burns via TON Connect, including transfer-with-comment and burn workflows.
  * Added NFT transfer via TON Connect, plus marketplace listing (GetGems) and purchase verification and buy instructions.
  * Included React and TypeScript end-to-end examples showing transaction assembly, valid-until handling, and payload/amount composition.

### DIFF
--- a/standard/tokens/jettons/transfer.mdx
+++ b/standard/tokens/jettons/transfer.mdx
@@ -138,6 +138,8 @@ To send jettons from a dApp through [TON Connect](/ecosystem/ton-connect/overvie
   import TonConnectUI from '@tonconnect/ui';
   import { beginCell, toNano, Address } from '@ton/ton';
 
+  const tonConnectUI = new TonConnectUI({ manifestUrl: '<MANIFEST_URL>' });
+
   const body = beginCell()
     .storeUint(0x0f8a7ea5, 32)
     .storeUint(0, 64)
@@ -170,6 +172,7 @@ Where:
 - `<DESTINATION_WALLET>` — the recipient's TON wallet address.
 - `<SENDER_WALLET>` — the sender's TON wallet address (receives excess fees).
 - `<SENDER_JETTON_WALLET>` — the sender's jetton wallet contract address. To find it, call the `get_wallet_address` get method on the jetton master contract, passing the sender's wallet address. See the [low-level example](#transfer-using-a-wallet-contract) below for the full derivation.
+- `<MANIFEST_URL>` — URL of the dApp manifest. See [manifest](/ecosystem/ton-connect/manifest).
 
 <Aside>
   `toNano` assumes 9 decimals. For tokens with different decimals (for example, 6 for USDT), compute the amount manually: `BigInt(value) * 10n ** BigInt(jettonDecimals)`.
@@ -245,6 +248,8 @@ The burn body follows [TEP-74 burn](https://github.com/ton-blockchain/TEPs/blob/
   ```ts title="TypeScript" icon="globe"
   import TonConnectUI from '@tonconnect/ui';
   import { beginCell, toNano, Address } from '@ton/ton';
+
+  const tonConnectUI = new TonConnectUI({ manifestUrl: '<MANIFEST_URL>' });
 
   const body = beginCell()
     .storeUint(0x595f07bc, 32)
--- a/standard/tokens/jettons/transfer.mdx
+++ b/standard/tokens/jettons/transfer.mdx
@@ -86,6 +86,196 @@ async function main() {
 void main();
 ```
 
+## Transfer via TON Connect
+
+To send jettons from a dApp through [TON Connect](/ecosystem/ton-connect/overview), build the [TEP-74](https://github.com/ton-blockchain/TEPs/blob/master/text/0074-jettons-standard.md#1-transfer) transfer body and submit it to the sender's jetton wallet contract:
+
+<CodeGroup>
+  ```tsx title="React" icon="react"
+  import { useTonConnectUI } from '@tonconnect/ui-react';
+  import { beginCell, toNano, Address } from '@ton/ton';
+
+  // transfer#0f8a7ea5 query_id:uint64 amount:(VarUInteger 16) destination:MsgAddress
+  // response_destination:MsgAddress custom_payload:(Maybe ^Cell)
+  // forward_ton_amount:(VarUInteger 16) forward_payload:(Either Cell ^Cell)
+  // = InternalMsgBody;
+
+  const body = beginCell()
+    .storeUint(0x0f8a7ea5, 32)                            // jetton transfer op code
+    .storeUint(0, 64)                                     // query_id:uint64
+    .storeCoins(toNano('0.001'))                          // amount — decimals vary per token (6 for USDT, 9 default)
+    .storeAddress(Address.parse('<DESTINATION_WALLET>'))   // destination:MsgAddress
+    .storeAddress(Address.parse('<SENDER_WALLET>'))        // response_destination:MsgAddress
+    .storeUint(0, 1)                                      // custom_payload:(Maybe ^Cell)
+    .storeCoins(toNano('0.05'))                           // forward_ton_amount — if >0, sends notification
+    .storeBit(1)                                          // forward_payload as reference
+    .storeRef(beginCell().endCell())                      // empty payload reference
+    .endCell();
+
+  const transaction = {
+    validUntil: Math.floor(Date.now() / 1000) + 360,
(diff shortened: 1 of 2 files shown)


standard/tokens/nft/transfer.mdx
```diff
@@ -73,3 +73,259 @@ where:
...
+
+  const marketplaceAddress = Address.parse(
+    'EQBYTuYbLf8INxFtD8tQeNk5ZLy-nAX9ahQbG_yl1qQ-GEMS',
+  ); // GetGems marketplace
+  const marketplaceFeeAddress = Address.parse(
+    'EQCjk1hh952vWaE9bRguFkAhDAL5jj3xj9p0uPWrFBq_GEMS',
+  ); // GetGems fee address
+  const destinationAddress = Address.parse(
+    'EQAIFunALREOeQ99syMbO6sSzM_Fa1RsPD5TBoS0qVeKQ-AR',
+  ); // GetGems sale deployer
```
delovoyhomie: `destinationAddress` is hardcoded to `EQAIFunALREOeQ99syMbO6sSzM_Fa1RsPD5TBoS0qVeKQ-AR`, but the current https://github.com/getgems-io/nft-contracts/blob/7f2cf76ea4dbe6b84da99796adc8fcb80d5a7e4e/readme.md?plain=1#L23-L24 explicitly says: "Do not use ...EQAIFun... to put NFTs for sale." Since this page is presented as a live how-to, I don't think we should keep a concrete deployer address that upstream now marks unsafe/deprecated. Could we replace this flow with the currently supported listing path, or remove the hardcoded address entirely?

standard/tokens/nft/transfer.mdx
```diff
@@ -73,3 +73,259 @@ where:
...
+      },
+    ],
+  };
+
+  const result = await tonConnectUI.sendTransaction(transaction);
+  ```
+</CodeGroup>
+
+Where `<NFT_SALE_CONTRACT>` is the address of the GetGems sale contract for the desired NFT item.
+
```
delovoyhomie, line 421: I think this needs a stronger safety note than just "use the desired sale contract address". The current GetGems README warns that direct on-chain purchases require checking the sale contract code and the correctness of the sale data; otherwise users can be tricked by lookalike contracts. Could we add an explicit verification step/caution here (for example, verify the sale contract code hash and sale data before sending TON), especially since this section is about buying directly without marketplace UI?

delovoyhomie (changes requested):
I found two safety-critical issues in the new GetGems section.

One example still hardcodes the old `EQAIFun...` deployer address from the deprecated cookbook, while the current upstream `getgems-io/nft-contracts` README explicitly warns not to use that contract for listing NFTs. The buy section also needs an explicit verification step before readers send TON directly to a sale contract.

Because these examples move real funds, I think both points should be fixed before merge.

standard/tokens/jettons/transfer.mdx
```diff
@@ -86,6 +86,201 @@ async function main() {
...
+    .storeBit(1)                                          // forward_payload as reference
+    .storeRef(beginCell().endCell())                      // empty payload reference
+    .endCell();
+
+  const transaction = {
+    validUntil: Math.floor(Date.now() / 1000) + 360,
+    messages: [
+      {
+        address: '<SENDER_JETTON_WALLET>',                // sender's jetton wallet address
+        amount: toNano('0.05').toString(),                // for gas fees, excess is returned
```
delovoyhomie, line 120: the jetton wallet needs `message_value > jetton_wallet_gas + forward_ton_amount + receiver_gas`, but here `forward_ton_amount = message value = 0.05 TON` so by the time the chain reaches the receiver's jetton wallet there's nothing left to forward as the notification. typical safe values are total `~0.1 TON` with forward `~0.02 TON` (matches the low-level example below at lines 352/360). this also applies to the bare TS variant on lines 150/160

correct me if i am wrong

```suggestion
    .storeUint(0, 1)                                      // custom_payload:(Maybe ^Cell)
    .storeCoins(toNano('0.02'))                           // forward_ton_amount — must be < message value
    .storeBit(0)                                          // forward_payload inline, empty
    .endCell();

  const transaction = {
    validUntil: Math.floor(Date.now() / 1000) + 360,
    messages: [
      {
        address: '<SENDER_JETTON_WALLET>',                // sender's jetton wallet address
        amount: toNano('0.1').toString(),                 // for gas fees, excess is returned
```

standard/tokens/jettons/transfer.mdx
```diff
@@ -86,6 +86,201 @@ async function main() {
...
+
+Where:
+
+- `<DESTINATION_WALLET>` — the recipient's TON wallet address.
+- `<SENDER_WALLET>` — the sender's TON wallet address (receives excess fees).
+- `<SENDER_JETTON_WALLET>` — the sender's jetton wallet contract address. To find it, call the `get_wallet_address` get method on the jetton master contract, passing the sender's wallet address. See the [low-level example](#transfer-using-a-wallet-contract) below for the full derivation.
+- `<MANIFEST_URL>` — URL of the dApp manifest. See [manifest](/ecosystem/ton-connect/manifest).
+
+<Aside>
+  `toNano` assumes 9 decimals. For tokens with different decimals (for example, 6 for USDT), compute the amount manually: `BigInt(value) * 10n ** BigInt(jettonDecimals)`.
```
delovoyhomie, line 178: tiny: the `Aside` has no `type=` so it defaults to a note. that's probably what's intended here, but explicit type makes it consistent with the rest of the file (lines 9, 26, etc)

```suggestion
<Aside type="note">
  `toNano` assumes 9 decimals. For tokens with different decimals (for example, 6 for USDT), compute the amount manually: `BigInt(value) * 10n ** BigInt(jettonDecimals)`.
</Aside>
```

standard/tokens/nft/transfer.mdx
```diff
@@ -73,3 +73,354 @@ where:
...
+import {
+  Address,
+  beginCell,
+  Cell,
+  contractAddress,
+  internal,
+  SendMode,
+  storeStateInit,
+  toNano,
+} from '@ton/ton';
```
delovoyhomie, line 173: `SendMode` is imported but never used in this snippet

standard/tokens/nft/transfer.mdx
```diff
@@ -73,3 +73,354 @@ where:
...
+interface ListingParams {
+  seller: Address;
+  nftAddress: Address;
+  royaltyAddress: Address;
+  royaltyPercent: number;
+  // Sale price in nanoTON
+  fullPrice: bigint;
+}
+
+const GETGEMS_FEE_PERCENT = 5000;
```
delovoyhomie, line 193: makes sense to spell out the unit? `5000` could be basis points (50%), per-mille (5%), or hundredths-of-percent — a reader has to dig into the contract source to find out. a single-line comment is enough

```suggestion
// GetGems platform fee, in 1/100000 (so 5000 = 5%)
const GETGEMS_FEE_PERCENT = 5000;
```

standard/tokens/nft/transfer.mdx
```diff
@@ -73,3 +73,354 @@ where:
...
+Once verification passes, send a payload-free transfer:
+
+<CodeGroup>
+  ```tsx title="React" icon="react"
+  import { useTonConnectUI } from '@tonconnect/ui-react';
+  import { toNano } from '@ton/ton';
+
+  // NFT price in TON; total value is price + 0.1 TON for network gas, with excess returned
+  const nftPrice = 5.0;
+  const buyAmount = nftPrice + 0.1;
```
delovoyhomie, line 374: `nftPrice + 0.1` happens in JS floats. `5.0 + 0.1` is fine but the moment `nftPrice` is, say, `0.2`, `0.2 + 0.1 = 0.30000000000000004` and `toNano` rejects it / rounds. safer to keep everything in nanoTON

```suggestion
  // NFT price in TON; total value is price + 0.1 TON for network gas, with excess returned
  const nftPrice = toNano('5.0');
  const buyAmount = nftPrice + toNano('0.1');
```

and drop the `.toString()`/`toNano(...)` around `buyAmount` below since it's already a `bigint`. same fix in the TS variant at lines 403-404

nabso45:
Nabso gidi from Ethiopia

Issue #30, opened 2025-09-03
Contract development


verytactical:
How to develop new good contracts.

Issue #29, opened 2025-09-03
Foundations


verytactical:
Blockchain foundations. Whitepapers explained for layman. Should contain everything there was in whitepapers.

PR #2083, opened 2026-04-14
feat(foundations): update config for [04.2026] release

### LINKED ISSUE #2074: [Foundations > Config and limits] Update config for [04.2026] release
Scope:
- Add documentation for Config 30
- Consensus_config_v4 changed

Extra reference: 
- https://github.com/ton-blockchain/ton/blob/af252bcdaea357fee21739e984654c2c84e7d61d/crypto/block/block.tlb#L782
- https://github.com/ton-blockchain/ton/releases/tag/v2026.04

reveloper:
Closes https://github.com/ton-org/docs/issues/2074.

## Summary by CodeRabbit

* **Documentation**
  * Expanded Param 29 docs to note its constructor evolution and new consensus fields, and added on-chain TL-B schema for multiple consensus config versions.
  * Added full Param 30: consensus extension section (gated by v2026.03 testnet / v2026.04 mainnet), documenting simplex config variants, fallback semantics to pre-2.0 catchain config, and an extensible noncritical_params map with encoding rules.
  * Included a table of recognized noncritical param IDs (0–14) and mainnet defaults (only 0, 1, 13 set).

[Review Change Stack](https://app.coderabbit.ai/change-stack/ton-org/docs/pull/2083?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


novusnota (commented):
fact-checking...

foundations/config.mdx
```diff
@@ -462,6 +462,99 @@ For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/
...
+  slots_per_leader_window:uint32
+  noncritical_params:(HashmapE 8 uint32)
+  = NewConsensusConfig;
+```
+
+### Configuration parameters of `simplex_config_v2`
+
+- `flags`: A general field that can be used to set various binary parameters.
+- `use_quic`: Whether the QUIC transport is used instead of RLDP2. Set to **True** in mainnet.
+- `slots_per_leader_window`: Number of consecutive slots assigned to one leader. Set to **4** in mainnet.
```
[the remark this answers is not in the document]
novusnota: Nope, it's correct, the TL-B's boolean true is capitalized: `True`

foundations/config.mdx
```diff
@@ -462,6 +462,99 @@ For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/
...
+| `8`  | `standstill_timeout`                   | `uint32` milliseconds      | `10000 ms`             | No-progress timeout before standstill recovery or rebroadcast logic triggers.                              |
+| `9`  | `standstill_max_egress_bytes_per_s`    | `uint32` number            | `6553600` (`50 << 17`) | Egress rate cap used during standstill rebroadcast.                                                        |
+| `10` | `max_leader_window_desync`             | `uint32`    number         | `250`                  | Maximum tolerated future leader-window distance for inbound Simplex traffic.                               |
+| `11` | `bad_signature_ban_duration`           | `uint32` milliseconds      | `5000 ms`              | Temporary ban duration after receiving bad signatures from a peer.                                         |
+| `12` | `candidate_resolve_rate_limit`         | `uint32`    number         | `10`                   | Per-peer rate limit for candidate resolution requests.                                                     |
```
novusnota:
```suggestion
| `10` | `max_leader_window_desync`             | `uint32` number            | `250`                  | Maximum tolerated future leader-window distance for inbound Simplex traffic.                               |
| `11` | `bad_signature_ban_duration`           | `uint32` milliseconds      | `5000 ms`              | Temporary ban duration after receiving bad signatures from a peer.                                         |
| `12` | `candidate_resolve_rate_limit`         | `uint32` number            | `10`                   | Per-peer rate limit for candidate resolution requests.                                                     |
```

foundations/config.mdx
```diff
@@ -462,6 +462,99 @@ For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/
...
+| `0`  | `target_rate`                          | `uint32` milliseconds      | `2400 ms`              | Target slot/block interval; used for leader pacing, block production timing, and skip scheduling.          |
+| `1`  | `first_block_timeout`                  | `uint32` milliseconds      | `1000 ms`              | Base timeout before skip voting starts for the first missing block in a leader window.                     |
+| `2`  | `first_block_timeout_multiplier`       | `float32` bits in `uint32` | `1.2`                  | Multiplier applied to `first_block_timeout` after a window that had skips.                                 |
+| `3`  | `first_block_timeout_cap`              | `uint32` milliseconds      | `100000 ms`            | Cap for the adaptive `first_block_timeout` growth.                                                         |
+| `4`  | `candidate_resolve_timeout`            | `uint32` milliseconds      | `1000 ms`              | Initial timeout for candidate / notarization resolution requests.                                          |
```
novusnota:
```suggestion
| `0`  | `target_rate`                          | `uint32` milliseconds      | `2400 ms`              | Target slot or block interval; used for leader pacing, block production timing, and skip scheduling.       |
| `1`  | `first_block_timeout`                  | `uint32` milliseconds      | `1000 ms`              | Base timeout before skip voting starts for the first missing block in a leader window.                     |
| `2`  | `first_block_timeout_multiplier`       | `float32` bits in `uint32` | `1.2`                  | Multiplier applied to `first_block_timeout` after a window that had skips.                                 |
| `3`  | `first_block_timeout_cap`              | `uint32` milliseconds      | `100000 ms`            | Cap for the adaptive `first_block_timeout` growth.                                                         |
| `4`  | `candidate_resolve_timeout`            | `uint32` milliseconds      | `1000 ms`              | Initial timeout for candidate or notarization resolution requests.                                         |
```

foundations/config.mdx
```diff
@@ -462,6 +462,99 @@ For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/
 
 [Parameter #29 on mainnet](https://tonviewer.com/config#29)
 
+## Param 30: consensus extension
+
+<Aside type="note">
+  The [TON v2026.03](https://github.com/ton-blockchain/ton/releases/tag/v2026.03) update introduced this parameter on testnet. With the [TON `v2026.04`](https://github.com/ton-blockchain/ton/releases/tag/v2026.04) update, `ConfigParam 30` was enabled on mainnet.
```
delovoyhomie: Small nit inside the Aside: `TON v2026.03` is plain text while `TON \`v2026.04\`` is wrapped in a code span, even though they are the same kind of release tag. Could we pick one and apply it to both? I would lean toward code spans for both, since the surrounding section refers to `ConfigParam 30` / `ConfigParam 29` in code font.

foundations/config.mdx
```diff
@@ -462,6 +462,99 @@ For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/
...
+</Aside>
+
+This parameter stores optional per-workchain settings for the [Simplex](https://github.com/ton-blockchain/simplex-docs/blob/main/Simplex.md)-based Catchain 2.0 consensus path. It does not fully replace `ConfigParam 29`: the node still uses `max_block_size` and `max_collated_data_size` from the `ConfigParam 29`.
```
delovoyhomie: Could we replace `ConfigParam 29` here with a deep link to the existing `## Param 29: consensus config` section on this same page (`[Param 29](#param-29-consensus-config)`)? The style guide asks for deep-linking to the exact section that contains the needed fact, and every other reference to a parameter on this page uses the `Param N` form — `ConfigParam N` is only used inside TL-B. Also the phrase `from the \`ConfigParam 29\`` at the end of the sentence reads a bit awkward; `from [Param 29](#param-29-consensus-config)` would fix both in one go.

foundations/config.mdx
```diff
@@ -462,6 +462,99 @@ For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/
...
+
+The `NewConsensusConfig` has two constructors:
+
+- `simplex_config#21` is the legacy fixed-layout format; scheduled for removal.
+- `simplex_config_v2#22` is the current extensible format, which supports arbitrary `noncritical_params` without changes to the `block.tlb` layout.
+
+The `simplex_config_v2#22` constructor moves noncritical configuration parameters into a sparse dictionary:
+
+```tlb
+simplex_config_v2#22 flags:(## 7)
```
delovoyhomie: This `simplex_config_v2#22` block is an exact duplicate of the one in the full schema above (lines 484–488). Could we drop this second copy? The sentence introducing it (`moves noncritical configuration parameters into a sparse dictionary`) already makes the point, and the table of IDs below makes the payload concrete — having the same TL-B twice on the same screen just invites the two copies to drift later if one is edited.

foundations/config.mdx
```diff
@@ -462,6 +462,99 @@ For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/
...
+  noncritical_params:(HashmapE 8 uint32)
+  = NewConsensusConfig;
+```
+
+### Configuration parameters of `simplex_config_v2`
+
+- `flags`: A general field that can be used to set various binary parameters.
+- `use_quic`: Whether the QUIC transport is used instead of RLDP2. Set to **True** in mainnet.
+- `slots_per_leader_window`: Number of consecutive slots assigned to one leader. Set to **4** in mainnet.
+- `max_block_size = 4 MiB`
```
delovoyhomie: According to the `simplex_config_v2#22` TL-B above, the constructor only has `flags`, `use_quic`, `slots_per_leader_window`, and `noncritical_params` — `max_block_size` and `max_collated_data_size` are not fields of this structure. The opening paragraph already says the node keeps reading those two from `ConfigParam 29`, so listing them under `### Configuration parameters of simplex_config_v2` reads as if they were part of the v2 constructor. Could we move these two bullets out of this list (or put them under a separate `Inherited from Param 29` sub-bullet) so the list matches the TL-B?

foundations/config.mdx
```diff
@@ -462,6 +462,99 @@ For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/
...
+| ---- | -------------------------------------- | -------------------------- | ---------------------- | ---------------------------------------------------------------------------------------------------------- |
+| `0`  | `target_rate`                          | `uint32` milliseconds      | `2400 ms`              | Target slot or block interval; used for leader pacing, block production timing, and skip scheduling.       |
+| `1`  | `first_block_timeout`                  | `uint32` milliseconds      | `1000 ms`              | Base timeout before skip voting starts for the first missing block in a leader window.                     |
+| `2`  | `first_block_timeout_multiplier`       | `float32` bits in `uint32` | `1.2`                  | Multiplier applied to `first_block_timeout` after a window that had skips.                                 |
+| `3`  | `first_block_timeout_cap`              | `uint32` milliseconds      | `100000 ms`            | Cap for the adaptive `first_block_timeout` growth.                                                         |
```
delovoyhomie: Style-guide nit ([`style-guide-extended.mdx`](https://github.com/ton-org/docs/blob/main/contribute/style-guide-extended.mdx) asks for thousands separators on values ≥ 10,000): in the `Default` column that applies to ID `3` (`100000 ms` → `100,000 ms`), IDs `6` and `8` (`10000 ms` → `10,000 ms`), ID `9` (`6553600` → `6,553,600`), and ID `14` (`15000 ms` → `15,000 ms`). Values below 10,000 (`2400`, `1000`, `5000`, etc.) are already fine.

foundations/config.mdx
```diff
@@ -462,6 +462,99 @@ For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/
...
+| `10` | `max_leader_window_desync`             | `uint32` number            | `250`                  | Maximum tolerated future leader-window distance for inbound Simplex traffic.                               |
+| `11` | `bad_signature_ban_duration`           | `uint32` milliseconds      | `5000 ms`              | Temporary ban duration after receiving bad signatures from a peer.                                         |
+| `12` | `candidate_resolve_rate_limit`         | `uint32` number            | `10`                   | Per-peer rate limit for candidate resolution requests.                                                     |
+| `13` | `min_block_interval`                   | `uint32` milliseconds      | `0 ms`                 | Minimum interval between parent block time and the next locally generated block.                           |
+| `14` | `no_empty_blocks_on_error_timeout`     | `uint32` milliseconds      | `15000 ms`             | How long empty-block fallback is allowed after the last finalized block when collation fails or times out. |
```
delovoyhomie: Every other parameter section on this page ends with a `[Parameter #N on mainnet](https://tonviewer.com/config#N)` link right before the next `##` heading (see `Param 29` at line 463 and `Param 31` at line 562). `Param 30` is the only section without it — could we add `[Parameter #30 on mainnet](https://tonviewer.com/config#30)` here for consistency?

delovoyhomie (commented):
Nice addition — the TL-B + `noncritical_params` mapping and the per-ID defaults table are exactly what was missing. A few inline notes: the main ones are that `max_block_size` / `max_collated_data_size` are listed as fields of `simplex_config_v2` but aren't actually in its TL-B, and that the `simplex_config_v2#22` code block is duplicated. Plus a few consistency things (missing `Parameter #30 on mainnet` footer link, `ConfigParam 29` not deep-linked to the on-page section, thousands separators).

reveloper:
Hello, @delovoyhomie. Thank you for the review! I'll be back to editing ~next week, I'm on my way to confirm some extra info for Consensus_v4.

reveloper:
> `ConfigParam 29` not deep-linked to the on-page section.

Hey, I asked Tonviewer devs if they can add this soon, so I'll deeplink this as well.

foundations/config.mdx
reveloper: WIP

reveloper:
@delovoyhomie, ready for review.

foundations/config.mdx
```diff
@@ -458,10 +464,137 @@ This parameter provides the configuration for the consensus protocol above `Catc
...
+The `NewConsensusConfig` has two constructors:
+
+- `simplex_config#21` is the legacy fixed-layout format; scheduled for removal.
+- `simplex_config_v2#22` is the current extensible format, which supports arbitrary `noncritical_params` without changes to the `block.tlb` layout.
+
+### Configuration parameters of `simplex_config_v2`
+
+- `flags`: A general field that can be used to set various binary parameters.
+- `use_quic`: Whether the QUIC transport is used instead of RLDP2. Set to **True** in mainnet.
+- `slots_per_leader_window`: Number of consecutive slots assigned to one leader. Set to **4** in mainnet.
```
delovoyhomie: bolding values like `**True**` and `**4**` falls under the no-bold-on-tokens rule of the style guide (see [emphasis](https://docs.ton.org/contribute/style-guide-extended#62-quotation-marks-and-emphasis)), code font keeps the literal grep-friendly and matches how `Bool` values are written elsewhere in the repo (lowercase)

```suggestion
- `use_quic`: Whether the QUIC transport is used instead of RLDP2. Set to `true` on mainnet.
- `slots_per_leader_window`: Number of consecutive slots assigned to one leader. Set to `4` on mainnet.
```

foundations/config.mdx
```diff
@@ -458,10 +464,137 @@ This parameter provides the configuration for the consensus protocol above `Catc
...
+- `noncritical_params`: A `HashmapE 8 uint32` map from parameter IDs to raw 32-bit values.
+
+Inherited from [Param 29](https://tonviewer.com/config#29):
+
+- `max_block_bytes` — maximum block size.
+- `max_collated_bytes` — maximum size of serialized block correctness proofs.
+
+The `noncritical_params` dictionary contains tunable timing and DoS-protection parameters that might differ between validators:
```
delovoyhomie: tiny but the phrasing `might differ between validators` reads as if every validator runs its own noncritical config, which isn't the case - the dictionary lives in `ConfigParam 30` on-chain, so all validators apply the same values. could we soften to something like `tunable across config updates without changing the block.tlb layout`?

```suggestion
The `noncritical_params` dictionary contains tunable timing and DoS-protection parameters that can be changed via a config update without altering the `block.tlb` layout:
```

foundations/config.mdx
```diff
@@ -458,10 +464,137 @@ This parameter provides the configuration for the consensus protocol above `Catc
...
+- `simplex_config_v2#22` is the current extensible format, which supports arbitrary `noncritical_params` without changes to the `block.tlb` layout.
+
+### Configuration parameters of `simplex_config_v2`
+
+- `flags`: A general field that can be used to set various binary parameters.
+- `use_quic`: Whether the QUIC transport is used instead of RLDP2. Set to **True** in mainnet.
+- `slots_per_leader_window`: Number of consecutive slots assigned to one leader. Set to **4** in mainnet.
+- `noncritical_params`: A `HashmapE 8 uint32` map from parameter IDs to raw 32-bit values.
+
+Inherited from [Param 29](https://tonviewer.com/config#29):
```
delovoyhomie: this link points at tonviewer's `#29` anchor, but in the param 30 intro (line 515) the same reference uses the internal anchor `[ConfigParam 29](/foundations/config#param-29-consensus-config)`

should we keep them consistent? @reveloper the internal anchor is also what the style guide prefers for cross-page references (see [links](https://docs.ton.org/contribute/style-guide-extended#122-what-to-link-and-what-not))

```suggestion
Inherited from [Param 29](/foundations/config#param-29-consensus-config):
```

foundations/config.mdx
```diff
@@ -458,10 +464,137 @@ This parameter provides the configuration for the consensus protocol above `Catc
...
+  proto_version:uint16 catchain_max_blocks_coeff:uint32 = ConsensusConfig;
+
+_ ConsensusConfig = ConfigParam 29;
+```
+
+<Aside>
+  `consensus_config_v4#d9` was introduced together with the [Catchain 2.0 / Simplex](https://github.com/ton-blockchain/simplex-docs/blob/main/Simplex.md) migration tracked by [Param 30](#param-30-consensus-extension). The `use_quic` toggle in Param 29 controls the transport for the legacy Catchain path; the `use_quic` toggle in Param 30 controls the transport for the new Simplex path. They are configured independently.
+</Aside>
 
 [Parameter #29 on mainnet](https://tonviewer.com/config#29)
```
delovoyhomie, line 512: small dup: the same `Parameter #29 on mainnet` link already shows up at line 467 (`For up-to-date values, see [Parameter #29 on mainnet](...)`), so this trailing standalone link is redundant after the rewrite. probably it makes sense to drop one of them?

```suggestion
```

foundations/config.mdx
```diff
@@ -458,10 +464,137 @@ This parameter provides the configuration for the consensus protocol above `Catc
 
 - `catchain_max_blocks_coeff`: The coefficient limiting the rate of block generation in `Catchain`, [description](https://github.com/ton-blockchain/ton/blob/master/doc/catchain-dos.md).
 
-For up-to-date values, see [https://tonviewer.com/config](https://tonviewer.com/config).
+For up-to-date values, see [Parameter #29 on mainnet](https://tonviewer.com/config#29).
+
+#### On-chain schema
+
+`ConfigParam 29` is a tagged union: validators must accept any of the legacy constructors so old serialized configs keep parsing, but new on-chain values are written using the latest tag. The current set defined in [`block.tlb`](https://github.com/ton-blockchain/ton/blob/af252bcdaea357fee21739e984654c2c84e7d61d/crypto/block/block.tlb#L782) is:
```
delovoyhomie: this `block.tlb` reference pins a commit hash (`af252bcdae...`), but the param 30 section a few hundred lines down pins a tag (`v2026.04`). both are stable, mixing them in adjacent sections looks accidental. could we settle on one form - probably the tag since it matches the release-note Aside in param 30?

```suggestion
`ConfigParam 29` is a tagged union: validators must accept any of the legacy constructors so old serialized configs keep parsing, but new on-chain values are written using the latest tag. The current set defined in [`block.tlb`](https://github.com/ton-blockchain/ton/blob/v2026.04/crypto/block/block.tlb#L782) is:
```

and the same line 436 in the Aside above

foundations/config.mdx
```diff
@@ -458,10 +464,137 @@ This parameter provides the configuration for the consensus protocol above `Catc
...
+| `6`  | `candidate_resolve_timeout_cap`        | `uint32` milliseconds      | `10,000 ms`              | Cap for candidate resolution timeout growth.                                                               |
+| `7`  | `candidate_resolve_cooldown`           | `uint32` milliseconds      | `10 ms`                  | Cooldown between candidate resolution attempts.                                                            |
+| `8`  | `standstill_timeout`                   | `uint32` milliseconds      | `10,000 ms`              | No-progress timeout before standstill recovery or rebroadcast logic triggers.                              |
+| `9`  | `standstill_max_egress_bytes_per_s`    | `uint32` number            | `6,553,600` (`50 << 17`) | Egress rate cap used during standstill rebroadcast.                                                        |
+| `10` | `max_leader_window_desync`             | `uint32` number            | `250`                    | Maximum tolerated future leader-window distance for inbound Simplex traffic.                               |
```
delovoyhomie: tiny inconsistency in the `Stored as` column: rows for IDs 9, 10, 12 use `uint32 number` while the rest are `uint32 milliseconds` / `float32 bits in uint32`. `number` adds no info that `uint32` doesn't already carry

wdyt to drop the suffix on the dimensionless rows so the column reads `uint32` / `uint32 milliseconds` / `float32 bits in uint32`?

```suggestion
| `9`  | `standstill_max_egress_bytes_per_s`    | `uint32`                   | `6,553,600` (`50 << 17`) | Egress rate cap used during standstill rebroadcast.                                                        |
| `10` | `max_leader_window_desync`             | `uint32`                   | `250`                    | Maximum tolerated future leader-window distance for inbound Simplex traffic.                               |
```

foundations/config.mdx
```diff
@@ -458,10 +464,137 @@ This parameter provides the configuration for the consensus protocol above `Catc
...
+  slots_per_leader_window:uint32
+  first_block_timeout_ms:uint32
+  max_leader_window_desync:uint32
+  = NewConsensusConfig;
+
+simplex_config_v2#22 flags:(## 7)
+  use_quic:Bool
+  slots_per_leader_window:uint32
+  noncritical_params:(HashmapE 8 uint32)
+  = NewConsensusConfig;
```
delovoyhomie, line 538: small but worth double-checking against the upstream [`block.tlb` at `v2026.04` line 782](https://github.com/ton-blockchain/ton/blob/v2026.04/crypto/block/block.tlb#L782): `consensus_config_v4#d9` (line 492 here) has `flags:(## 6) { flags = 0 }`, while both `simplex_config#21` and `simplex_config_v2#22` have `flags:(## 7)` with no `{ flags = 0 }` constraint. is that intentional, or did the `flags = 0` predicate just get dropped in the copy? if the upstream schema does require `flags = 0` for the simplex configs too, we should keep it here for reviewers who use this page as the spec reference

foundations/config.mdx
reveloper: I think, in this context better to highlight parsed values. In the original text, it was so(they were just printed); that's why I added a direct link to the Tonviewer-parsed values.
For me, consistency here works like this, since the current value will not be visible with one click. 

Do you think we should redirect this to an internal anchor?

reveloper:
Thank you for the review. I'll address the comments during the next week.

foundations/config.mdx
reveloper: Thanks! Fair, updated.

foundations/config.mdx
reveloper: That's fair if we consider Parameter #29 inside, but is it fine to break the pattern for the entire config document, where you aligned the link as the last row in every single Parameter chapter?
Probably, during skimming, the user may expect to find this link at the end of the chapter, according to others.

The only thing is that I probably place such links for the entire document at the beginning of every Param chapter.

foundations/config.mdx
reveloper: There are not mistake here, this is correpsponds to block.tlb content.

In the `ConfigParam 29`:
```
consensus_config_new#d7 flags:(## 7) { flags = 0 }
```
In the `ConfigParam 30`:
```
simplex_config#21 flags:(## 7) use_quic:Bool
```

novusnota (approved):

PR #2085, opened 2026-04-15
chore(toncenter): hide generated API pages from search engines and AI contexts outside of docs MCP

### LINKED ISSUE #1925: [Ecosystem > APIs] Consider hiding generated API pages from search
Otherwise, they pollute the search and make it much less usable. This issue is only about the `.mdx` pages auto-generated based on corresponding routes in OpenAPI specification files that we have (mostly for TON Center).

As for the solution, it can be just to add a `noindex` field to the frontmatter. [Consult the documentation](https://www.mintlify.com/docs/organize/hidden-pages#understanding-hidden-versus-noindex).

Image

novusnota:
Closes #1925

We might want to keep certain pages in and leave out the others, but that'll hurt the consistency of the search. Not sure whether to proceed with this change or not, so let's postpone it for a while.

UPD: Change seems reasonable, since API pages are rarely searched that way.

## Release Notes

* **Documentation**
  * Updated numerous API documentation pages to exclude them from search engine indexing.

* **Chores**
  * Enhanced spell-check configuration with new rule overrides and pattern matching.
  * Added development environment directory to git ignore list.


novusnota:
@coderabbitai please, go take a look at https://www.mintlify.com/docs/llms.txt and check whether this change here will hurt AI assistant in the docs or not

I'd wager it'll hide exact API pages in favor of using v2/overview.mdx and v3/overview.mdx files for the reference.

PR #2091, opened 2026-04-16
fix(ai-reviews): try to correct maintainer recognition checks

### LINKED ISSUE #1970: Correct maintainer recognition checks in manual AI reviews
Currently, `/review` command does not take the "docs" team members into account, but it (probably) should.

novusnota:
Closes #1970

### DIFF
--- a/.github/workflows/pitaya.yml
+++ b/.github/workflows/pitaya.yml
@@ -45,7 +45,9 @@ jobs:
       cancel-in-progress: true
     # Run on:
     # - PR events when ready_for_review or opened as non‑draft
-    # - Issue comments only when it's a PR thread, command is /review, and commenter is trusted
+    # - Issue comments only when it's a PR thread and command is /review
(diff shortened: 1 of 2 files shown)


novusnota:
@coderabbitai suggest alternatives: perhaps, it's better to first collect all info on the user asking for `/review`, then proceed with the separate job for handling it? What if commander.yml is to be used instead? Outline your suggestions in a single message.

novusnota:
This should be closed — maintainer checks in the future would be based solely on repository access, as that is easier to manage wrt access rights than teams within an organization. 

Plus, future AI setup should primarily focus on checking factual technical accuracy and consistency/typos. The style guide adherence checks should become opt-in for folks who want them for their contributions, like me. Still, it should probably not run by default lest it become noisy and unwieldy. After all, the plan is for many teams to co-exist in the documentation, and agreeing on a single strict(est) style guide for all seems infeasible.

PR #2102, opened 2026-04-20
chore(zk): update TON verifier export commands

### LINKED ISSUE #2124: [Contract dev > Techniques > ZK] Update export commands

mysteryon88:
Closes #2124

## Summary

Updated the zero-knowledge verifier export documentation to match the latest `export-ton-verifier` interface.

The utility now uses Tolk as the default target, so the examples were adjusted to export Tolk contracts without an extra flag and to use `--func` explicitly when generating FunC contracts.

Since Tolk has continued to evolve and the generated verifier now uses newer language features that were not available before, the wrapper flow also had to be updated. The docs now distinguish the wrapper import commands for Tolk and FunC versions so users copy the correct TypeScript wrapper for the contract target they generate.

## Changes

- Updated verifier export examples to make Tolk the default output.
- Added explicit `--func` usage for FunC verifier generation.
- Updated wrapper import commands for Tolk and FunC.
- Reordered Arkworks and Gnark examples to reflect the current default flow.

## Summary by CodeRabbit

* **Documentation**
  * Clarified verifier export workflow with an explicit directory navigation step and refined command examples for different contract targets.
  * Rewrote wrapper instructions to describe copying and importing generated wrappers and updated guidance for test imports.
  * Reordered examples to present the new default-first flow and normalized examples across exporters.


novusnota (approved):
@mysteryon88 Thanks for the update!

@kay-is, I'll merge this one first so your PR (#2128) can adjust accordingly. Otherwise, merging yours first would cause too many conflicts here. Sorry for the slight inconvenience on your end :)

PR #2110, opened 2026-04-22
feat(contract-dev): edit upgrade contracts to fit styleguide

### LINKED ISSUE #2109: [Contract dev > Techniques > Upgrading contracts] Style cleanup
The upgrade techniques guide needs to be edited to fit the style guide.

kay-is:
closes #2109

## Summary by CodeRabbit

* **Documentation**
  * Clarified upgradeability narrative and when address preservation is required
  * Rewrote "How upgrades work" to distinguish code activation after transactions vs data replacement during execution
  * Added explicit guidance on insufficient-funds behavior and a caution that certain send options can leave a contract unusable
  * Converted delayed/hot sections into action-oriented "pattern" sections, revised flows, migration risk warnings, and updated examples and links


contract-dev/techniques/upgrades.mdx
```diff
@@ -30,27 +29,29 @@ Key difference: `setCodePostponed()` applies changes _after_ the current transac
...
+1. Send an upgrade message to the contract that contains new code, data, or both.
+1. Verify that the message comes from an admin address.
+1. If the message contains code, schedule the code replacement with `setCodePostponed()`.
+1. If the message contains data, replace the existing data with `setData()`.
+1. After the transaction completes, apply the new code during the action phase.
+1. Process subsequent messages with the new code.
 
-The upgrade happens immediately in a single transaction. The new code becomes active after the transaction completes. Any data replacement happens during the compute phase, so the new data is immediately available when the transaction ends.
+The upgrade runs in a single transaction. New code becomes active after the transaction completes, and new data is available when the transaction ends. If the message does not provide enough Toncoin to run both the compute phase and the action phase, the entire transaction is aborted and no state changes from the upgrade are applied. Test the upgrade script to estimate gas requirements, and send enough Toncoin to execute the full upgrade transaction.
```
delovoyhomie, line 50: Wondering about this one technically. The stdlib reference says `contract.setCodePostponed()` creates an output action that takes effect only after successful termination, while `contract.setData()` writes storage during the compute phase. If the compute phase succeeds (so `setData` commits) but the action phase then runs out of Toncoin, I'd expect the storage change to persist while the SETCODE action does not - leaving the contract with the new data schema but still the old code.

If that's right, then "the entire transaction is aborted and no state changes from the upgrade are applied" is misleading, and this is actually a fairly sharp footgun worth calling out directly. Could be worth verifying against the actual TVM semantics and either correcting the claim or adding a warning.

contract-dev/techniques/upgrades.mdx
```diff
@@ -96,25 +97,25 @@ fun Storage.save(self) {
 }
 ```
 
-## Delayed upgrades for production safety
-
-When upgrading protocols that are already running and have users, delayed upgrades are a best practice. This provides additional security layers: if an admin is compromised, there is time to react. Users can also see the upgrade and withdraw funds from the protocol if it has been compromised.
+## Delayed upgrade pattern for increased security
```
delovoyhomie: Small thing - the heading suffix `for increased security` feels a bit tacked on. The intro paragraph right below already explains the security motivation, so the heading could just be `Delayed upgrade pattern` and read a bit cleaner. Same thought for `Hot upgrade pattern for high update frequencies` below.

contract-dev/techniques/upgrades.mdx
```diff
@@ -215,36 +216,36 @@ fun Storage.save(self) {
...
+1. Send an upgrade message with the new code cell and optional additional data.
+1. Verify the message comes from an admin address.
+1. Call `setCodePostponed()` to schedule the code replacement.
+1. Call `setTvmRegisterC3()` to activate the new code in register [C3](/tvm/registers#c3-—-function-selector) immediately.
+1. Call `hotUpgradeData()` to run the migration with the new code.
 
-The key mechanism: `setTvmRegisterC3()` switches the code register so the migration function executes with the new code in the same transaction. The migration reads the current storage state (preserving all updates), transforms it to the new schema, and saves it. When the transaction completes, the new code becomes permanent through `setCodePostponed()`.
+The `setTvmRegisterC3()` is the key to hot upgrades. It replaces the current code immediately so the following command (e.g., `hotUpgradeData()`) executes the new code, which reads the current storage state (preserving all updates), transforms it to the new schema, and saves it. After the transaction completes, the new code becomes permanent through `setCodePostponed()`.
```
delovoyhomie: The `which reads ...` clause is a little ambiguous about its antecedent - it's not immediately obvious whether `which` refers to the new code, `hotUpgradeData()`, or the command in parentheses. Might read cleaner as two sentences, e.g.: `It replaces the current code immediately, so the next call (such as to hotUpgradeData()) runs under the new code. That migration function reads the current storage, transforms it to the new schema, and saves it.`

contract-dev/techniques/upgrades.mdx
```diff
@@ -411,35 +408,38 @@ fun Storage.save(self) {
 }
 ```
 
-In the new version, `hotUpgradeData()` performs the migration:
+The new version of `hotUpgradeData()` function is what is called after the code was switched with `setTvmRegisterC3()` and performs the migration.
```
delovoyhomie, line 419: (Posting here because line 349 is outside the diff.) Minor consistency thing in the `new.tolk` code block on line 349 - `struct oldStorage` is camelCase, but every other struct in this file is PascalCase (`Storage`, `HotUpgrade`, `CurrentRequest`, `UpgradeContract`, etc.). The Tolk convention is that lowercase struct names are reserved for empty namespace-style structs that group static methods (e.g., `struct blockchain` for `blockchain.now()`), see [Empty structures](https://github.com/ton-org/docs/blob/main/tolk/syntax/structures-fields.mdx#empty-structures). Since `oldStorage` has real fields and a `.load()` method, `OldStorage` would fit the convention and also avoid the variable-shadows-type situation on line 336 (`var oldStorage = lazy oldStorage.load();`).

contract-dev/techniques/upgrades.mdx
```diff
@@ -411,35 +408,38 @@ fun Storage.save(self) {
...
+
+The migration logic follows these steps:
 
-1. Loads storage using the old structure (`oldStorage` with `adminAddress` and `counter`)
-1. Creates new storage with the additional `metadata` field from `additionalData`
-1. Reorders fields (`counter` moves before `adminAddress`)
-1. Writes the migrated storage immediately with `contract.setData()`
+1. Load the storage using the old structure (e.g., `oldStorage` with `adminAddress` and `counter`).
```
delovoyhomie: Minor consistency thing: `oldStorage` is camelCase, but every other struct in this file is PascalCase (Storage, HotUpgrade, CurrentRequest, UpgradeContract, etc). Renaming to OldStorage would match the rest, and it'd also avoid the variable-shadows-type situation on line 336 (`var oldStorage = lazy oldStorage.load();`)

https://github.com/ton-org/docs/blob/main/tolk/syntax/structures-fields.mdx#empty-structures

delovoyhomie (changes requested):

contract-dev/techniques/upgrades.mdx
kay-is: Good point!

As far as [the docs explain it](https://docs.ton.org/foundations/phases): 

"If an action fails (for example, due to a lack of funds), the transaction may revert or skip the action, depending on its mode. For example, mode = 0, flag = 2 means that any errors arising while processing this message during the action phase are ignored."

Depending on these modes, it seems both can be true.

kay-is: It seems that by default, everything is reverted, but setting the `SendIgnoreErrors` flag for a message will ignore the insufficient funds error, and can lead to a situation where the storage is updated, but the code is not. 

I added a caution aside below to explain it, because it is possible, but not the default behavior.

contract-dev/techniques/upgrades.mdx
```diff
@@ -30,27 +29,37 @@ Key difference: `setCodePostponed()` applies changes _after_ the current transac
...
-  Use delayed upgrades to allow users to [react to compromised admin keys](https://blog.trailofbits.com/2025/06/25/maturing-your-smart-contracts-beyond-private-key-risk/) or [unwanted updates](https://en.wikipedia.org/wiki/Exit_scam).
+  Use delayed upgrades to allow users to react to compromised admin keys or unwanted updates.
+
+  [The Trail of Bits blog](https://blog.trailofbits.com/2025/06/25/maturing-your-smart-contracts-beyond-private-key-risk/) and [Wikipedia](https://en.wikipedia.org/wiki/Exit_scam) provide additional information on this topic.
 </Aside>
 
 ## Basic upgrade pattern
 
-The contract accepts upgrade messages containing new code and data. Only the admin can trigger upgrades.
+Upgradable contracts accept upgrade messages containing new code and data. Only an admin can trigger upgrades.
```
aigerimu: `Upgradable contracts` sounds like we need a definition of which contracts are upgradable and which are not, wdyt?

contract-dev/techniques/upgrades.mdx
```diff
@@ -30,27 +29,37 @@ Key difference: `setCodePostponed()` applies changes _after_ the current transac
...
+</Aside>
 
-If there's not enough Toncoin to execute action phase with the code update, but enough to execute compute phase with data update, the contract state changes will be reverted. Make sure to thoroughly test the upgrade script for possible gas issues, and provide enough Toncoin to execute the upgrade transaction completely.
+### Example contract
 
-### Example
+The following contract accepts `UpgradeContract` messages that contain new code or data. Only admins can trigger upgrades.
 
 ```tolk title="Tolk" expandable
 struct (0x1111) UpgradeContract {
```
delovoyhomie, line 65: Minor nit - `0x1111` is the `UpgradeContract` opcode tag here, and `throw 1111` (line 80, just below) is also the unauthorized-sender exit code. Reusing the same number for two different things makes failures slightly harder to trace from a tx dump. The delayed pattern below uses `100` for the same auth check; might be worth aligning

@Shvandre wdyt?

contract-dev/techniques/upgrades.mdx
```diff
@@ -96,25 +105,25 @@ fun Storage.save(self) {
...
-The admin can also send `RejectUpgrade` at any time to cancel a pending upgrade. This three-message flow (request → wait → approve or reject) gives users time to review changes and react if the admin account is compromised.
+Admins can also send `RejectUpgrade` at any time to cancel a pending upgrade. This three-message flow (request → wait → approve or reject) gives users time to review changes and react if an admin account is compromised.
 
-### Example
+### Example contract
+
+The following code illustrates the delayed upgrade pattern. The contract accepts `RequestUpgrade`, `RejectUpgrade`, and `ApproveUpgrade` messages. Only admins can trigger these actions.
 
 ```tolk title="Tolk" expandable
 struct UpgradeContract {
```
delovoyhomie, line 129: (re: `CurrentRequest` field on lines 134, 161, 163, 175, 177, 185-198, 209, 215 - anchored here at the closest in-diff line)

small consistency thing: `CurrentRequest` is used as both a struct type (line 134) and a struct field (`storage.CurrentRequest`)

Tolk convention is PascalCase for types and camelCase for fields/variables, so `currentRequest` for the field would line up with the rest of the file (`adminAddress`, `timeout`, `timestamp`, `newUpgrade`) and disambiguate the type/field references at a glance 

cc @Shvandre

contract-dev/techniques/upgrades.mdx
```diff
@@ -215,36 +224,36 @@ fun Storage.save(self) {
...
+### Example code
 
-The example shows a counter contract that adds a metadata field through a hot upgrade. The storage structure changes: the original version stores only `adminAddress` and `counter`. The new version adds `metadata` and reorders fields.
+The example shows a counter contract that changes the storage structure through a hot upgrade. The original version stores only `adminAddress` and `counter`; the new version adds `metadata` and reorders fields.
 
-Original contract:
+The original contract code before the upgrade:
 
 ```tolk title="main.tolk" expandable
 import "@stdlib/tvm-lowlevel"
```
delovoyhomie, line 259: line 274, `fun hotUpgradeData(additionalData: cell?) { return null; }`

this returns `null` from a function with no declared return type. Since the body is never actually executed (the new code's version takes over after `setTvmRegisterC3`), `return null` reads as if the function is supposed to produce a value. 

IMHO it will probably cleaner as:

```diff
 @method_id(2121)
-fun hotUpgradeData(additionalData: cell?) { return null; }
+fun hotUpgradeData(additionalData: cell?) { }
```

…optionally with a one-line comment noting that the body is intentionally empty because it's replaced before being called 

cc @Shvandre

delovoyhomie (dismissed):

contract-dev/techniques/upgrades.mdx
Shvandre: > Reusing the same number for two different things makes failures slightly harder to trace from a tx dump

Not really. But if you want to make those numbers different - it is also okay.

contract-dev/techniques/upgrades.mdx
Shvandre: Wow, I am not sure that this code compiles. If it compiles, let's use function with empty body.

contract-dev/techniques/upgrades.mdx
Shvandre: Yeah, Slava is 100% right

Shvandre (approved):
good

contract-dev/techniques/upgrades.mdx
albertoblue87-netizen: So we upload it in red? It doesn't matter, right?

contract-dev/techniques/upgrades.mdx
```diff
@@ -411,35 +408,38 @@ fun Storage.save(self) {
...
-You can combine delayed upgrades with hot upgrades for production protocols that require both safety and structure migration. The delayed pattern provides time for users to review changes, while the hot upgrade mechanism handles storage migration without data loss.
+Combine delayed upgrades with hot upgrades for production protocols that require both safety and structure migration. The delayed upgrade pattern provides time for users to review changes, while the hot upgrade mechanism handles storage migration without data loss.
 
 <Aside
-  title={"📁 Complete Example Code"}
+  title="Complete example code"
+  type="tip"
 >
-  You can find full working examples demonstrating all upgrade patterns in our [GitHub repository](https://github.com/ton-org/docs-examples/tree/main/contract-dev/Upgrading). This includes implementations for basic, delayed, and hot upgrade patterns.
+  The [TON Examples GitHub repository](https://github.com/ton-org/docs-examples/tree/main/contract-dev/Upgrading) contains full working examples demonstrating all upgrade patterns. This includes implementations for basic, delayed, and hot upgrade patterns.
```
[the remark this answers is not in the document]
albertoblue87-netizen: > ### [HIGH] Tip Aside title and link text could better describe benefit
> 
> Lines 440–444 define a tip aside with `title="Complete example code"` and link text “The [TON Examples GitHub repository]… contains full working examples…”. While functional, this combination is somewhat generic and does not fully describe the benefit of the linked resource, which is to provide concrete upgrade pattern implementations. The style guide prefers descriptive link text and concise callouts that clearly state what the reader gains.
> 
> Please leave a reaction 👍/👎 to this suggestion to improve future reviews for everyone!alberto Constantino

albertoblue87-netizen (commented):
Done

contract-dev/techniques/upgrades.mdx
```diff
@@ -30,27 +29,29 @@ Key difference: `setCodePostponed()` applies changes _after_ the current transac
...
-1. Code is scheduled. If new code is provided, `setCodePostponed()` schedules it for replacement.
-1. Data is upgraded. If new data is provided, `setData()` immediately replaces the storage.
-1. Transaction completes. The action phase executes, applying the new code.
-1. Next message uses new code. Subsequent messages execute with the upgraded logic.
+1. Send an upgrade message to the contract that contains new code, data, or both.
+1. Verify that the message comes from an admin address.
+1. If the message contains code, schedule the code replacement with `setCodePostponed()`.
+1. If the message contains data, replace the existing data with `setData()`.
+1. After the transaction completes, apply the new code during the action phase.
+1. Process subsequent messages with the new code.
```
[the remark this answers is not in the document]
albertoblue87-netizen: > ### [HIGH] Non-imperative steps and inconsistent list punctuation
> 
> The ordered list under “How it works” uses complete sentences with repeated subjects instead of concise, imperative instructions, which reduces scannability in a procedural section. Each item should be a direct action in imperative mood with a single, clear operation per step. The current wording also mixes descriptive phrasing with instructions, making it harder for readers to quickly follow the upgrade procedure. Aligning the steps with the style guide’s preference for present tense, active voice, and imperative mood will make the procedure clearer and easier to scan.
> 
> ``Alberto Const.

contract-dev/techniques/upgrades.mdx
```diff
@@ -411,35 +408,38 @@ fun Storage.save(self) {
...
 
-Use standard upgrades when:
+Use standard upgrades instead when:
 
-- The contract upgrades infrequently
-- You can predict storage state at upgrade time
-- Simpler upgrade logic reduces risk
+- The contract upgrades infrequently.
+- Storage state at upgrade time is predictable.
+- Simpler upgrade logic would reduce risk.
```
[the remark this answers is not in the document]
albertoblue87-netizen: > ### List items use inconsistent trailing periods
> 
> The bulleted lists under “When to use hot upgrades” and “Use standard upgrades instead when” use inconsistent terminal punctuation. Within each list, one item ends with a period while the others do not, violating the documentation style guideline that list items in the same list use consistent punctuation. This inconsistency slightly reduces visual clarity and makes the text appear less polished. The primary issue arises directly from the recently edited list items, which introduced periods on some bullets but not others.
> 
>this suggestion to improve future reviews for everyone!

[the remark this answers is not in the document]
albertoblue87-netizen: > ### [HIGH] List items use inconsistent trailing periods
> 
> The bulleted lists under “When to use hot upgrades” and “Use standard upgrades instead when” use inconsistent terminal punctuation. Within each list, one item ends with a period while the others do not, violating the documentation style guideline that list items in the same list use consistent punctuation. This inconsistency slightly reduces visual clarity and makes the text appear less polished. The primary issue arises directly from the recently edited list items, which introduced periods on some bullets but not others.
> 
> Please leave a reaction 👍/👎 to this suggestion to improve future reviews for everyone!

PR #2115, opened 2026-04-22
feat(contract-dev): edit random numbers guide to fit style guide and use Tolk code

### LINKED ISSUE #2113: [Contract dev > Techniques > Random numbers] Style cleanup & Tolk port
The random number doc should adhere to the style guide and use Tolk code for its examples.

kay-is:
Fixes #2113.

## Summary by CodeRabbit

* **Documentation**
  * Adopted TON Tolk RNG workflow: explicit initialization and integer/range retrieval.
  * Reorganized into a consolidated comparison table and clearer sections for single-block, multi-block skipping, and commit-reveal.
  * Expanded commit-reveal: commitments, verification, duplicate/mismatch rejection, deterministic transcript, and final hashing.
  * Updated security trade-offs, best practices, and testing guidance (testnet); presentation and caution notes refreshed.


contract-dev/techniques/random.mdx
```diff
@@ -5,164 +5,146 @@ sidebarTitle: "Random numbers"
...
+| Resistance to user manipulation      | High                                     | High                                                         | High                                                               |
+| Resistance to validator manipulation | Low                                      | Medium                                                       | High                                                               |
+| Cost (gas + storage)                 | Low                                      | Medium                                                       | High                                                               |
 
-The [`randomize_lt()`](/languages/func/stdlib#randomize-lt) function mixes the transaction's logical time into the random seed before generating random numbers. This provides basic entropy from blockchain state.
+<Aside
+  type="caution"
+>
+  When choosing an approach, consider the value at risk and required time-to-finality. For high-stakes applications such as lotteries with significant funds, use commit-reveal. Audit implementations through formal verification when possible.
+</Aside>
```
delovoyhomie, line 31: This callout reads as advisory ("consider…", "use commit-reveal", "audit when possible") rather than a hazard warning — I think `type="tip"` would fit better, which is what the original used

(the bot flagged "use commit-reveal" as second-person, but second-person imperative is explicitly allowed by the [style guide](https://docs.ton.org/contribute/style-guide) and common across these docs)

contract-dev/techniques/random.mdx
```diff
@@ -5,164 +5,146 @@ sidebarTitle: "Random numbers"
...
+The [`random.initialize()`](/tolk/features/standard-library#random-initialize) function initializes the random number generator with a seed derived from the transaction's logical time. This provides basic entropy from blockchain state.
 
 ### How it works
 
-Call [`randomize_lt()`](/languages/func/stdlib#randomize-lt) once before using [`random()`](/languages/func/stdlib#random) or [`rand()`](/languages/func/stdlib#rand) functions. The logical time adds variability to the seed, making it harder for observers to predict the outcome without executing the transaction.
+Call `random.initialize()` once before using [`random.uint256()`](/tolk/features/standard-library#random-uint256) or [`random.range()`](/tolk/features/standard-library#random-range-limit) functions. The logical time adds variability to the seed, making it harder for observers to predict the outcome without executing the transaction.
 
-```func
-randomize_lt();
+Code example:
```
delovoyhomie: I think the "Code example:" line can be dropped — the fenced block already has `title="Tolk"`, so the label is self-evident.

contract-dev/techniques/random.mdx
```diff
@@ -5,164 +5,146 @@ sidebarTitle: "Random numbers"
...
+The contract initiates an operation, then waits for responses that arrive several blocks later. The random seed from future blocks, which the initiating validator does not control, determines the result.
 
 ### Security model
 
-- ✅ Resistant to regular users
-- ✅ More resistant to single validators than `randomize_lt()`
-- ⚠️ Still vulnerable to determined validators who can coordinate timing across multiple blocks they generate
+Block skipping is resistant to attacks from regular users and more resistant to single validators than single-block randomness.
+
+However, it is still **vulnerable to determined validators** who represent 1/250 of the network, as they can still choose optimal timing to influence outcomes in blocks they generate.
```
delovoyhomie: small number-agreement issue: "determined validators who represent 1/250 of the network" reads as if each validator represents 1/250, but the 1/250 figure is about a *single* validator's share of the network. The original phrasing was clearer: `A validator representing 1/250 of the network can still choose optimal timing…`

also the validator count is a config parameter `ConfigParam 16` and changes over time, so any hard-coded fraction like 1/250 will drift. Might be cleaner to drop the exact fraction and say something like `a single validator out of the total validator set can still choose optimal timing…`

now max_validators is 400
https://tonviewer.com/config#:~:text=#,-16%20%2D%20Validators%20limits

contract-dev/techniques/random.mdx
```diff
@@ -5,164 +5,146 @@ sidebarTitle: "Random numbers"
...
 
 ### Security model
 
-- ✅ Cryptographically secure when properly implemented
-- ✅ Resistant to both user and validator manipulation; no single party can predict or influence the outcome
-- ⚠️ Requires incentives and penalties to prevent participants from refusing to reveal
-- ⚠️ Validators can influence timing or censor messages, but cannot determine the random value
+The commit-reveal scheme is the most secure on-chain randomness approach, as it relies on cryptographic commitments and multiple independent participants. It prevents any single participant from unilaterally determining the final value.
+
+Yet, **validators can still influence timing or censor messages**, but they cannot determine the final random value without colluding with participants.
```
delovoyhomie: imho "Yet," as a sentence opener reads a bit awkwardly here. I'd use `However,` or just drop the conjunction: `Validators can still influence timing or censor messages, but they cannot determine the final random value without colluding with participants.`

contract-dev/techniques/random.mdx
```diff
@@ -5,164 +5,146 @@ sidebarTitle: "Random numbers"
...
+- Decentralized gaming with financial stakes.
+- Systems requiring Byzantine fault tolerance.
 
 ## Best practices
 
-- Avoid standalone [`random()`](/languages/func/stdlib#random) calls. Validators controlling the block seed can predict the output.
-- Keep randomness out of external message receivers. External messages remain vulnerable even with [`randomize_lt()`](/languages/func/stdlib#randomize-lt).
-- Use hybrid or off-chain entropy for critical applications. Combine on-chain randomness with off-chain entropy or external randomness oracles when significant value is at risk.
-- Test randomness behavior across different blocks. Verify that the contract behaves correctly when randomness is manipulated within validator capabilities.
+- Always call `random.initialize()` before using `random.uint256()` or `random.range()` to prevent users from predicting random values.
```
delovoyhomie: I think this bullet understates the real hazard a bit. A reader could conclude that `random.initialize()` is sufficient for security, but the rest of the page makes clear it isn't (validators controlling the block seed can still predict outputs).

The original framing was sharper: `Avoid standalone random() calls. Validators controlling the block seed can predict the output.`

Maybe restore that, or at least hedge: `Always call random.initialize() before using random.uint256() or random.range(); this prevents users from predicting values but does not protect against validators who control the block seed.`?

contract-dev/techniques/random.mdx
```diff
@@ -176,19 +158,23 @@ The block seed is not used directly in contracts. Instead, it is [hashed with th
...
 ### Random number generation
 
-The [`RANDU256`](/tvm/instructions#f810-randu256) TVM instruction implements the actual random number generation. When called:
+The [`RANDU256`](/tvm/instructions#f810-randu256) TVM instruction implements the actual random number generation.
 
-1. Take the current seed `r` (32 bytes)
-1. Compute `SHA512(r)`
-1. Use the first 32 bytes as the new seed
-1. Return the remaining 32 bytes as the random number
+The process is as follows:
```
delovoyhomie: small thing here: I think the "The process is as follows:" line can just be dropped. The numbered list below introduces itself. wdyt?

delovoyhomie (dismissed):

contract-dev/techniques/random.mdx
```diff
@@ -5,164 +5,146 @@ sidebarTitle: "Random numbers"
...
+Multiple participants commit to secret values by publishing hashes, then reveal those values in a later phase. The final randomness is derived from combining all revealed values, ensuring no single party can determine the outcome alone. When properly implemented, this approach provides cryptographically secure randomness suitable for high-value applications.
 
 ### How it works
 
-1. Commit phase: Each participant generates a random number off-chain and submits its hash to the contract.
-1. Reveal phase: After all commitments are received, participants disclose their original numbers.
-1. Combination: The contract combines the revealed numbers (e.g., by XOR or sum) to produce the final random value.
+1. **Commit phase**: Each participant generates a random number off-chain and submits its hash to the contract.
+1. **Reveal phase**: After all commitments are received, participants disclose their original numbers.
+1. **Combination**: The contract combines the revealed numbers (e.g., by XOR or sum) to produce the final random value.
```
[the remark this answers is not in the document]
kay-is: Hey @novusnota or @delovoyhomie, can one of you confirm this?

It sounds reasonable, but my crypto skills aren't strong enough to check this without substantial effort 🫣

contract-dev/techniques/random.mdx
```diff
@@ -5,164 +5,142 @@ sidebarTitle: "Random numbers"
...
+| Cost (gas + storage)                 | Low                                      | Medium                                                       | High                                                               |
 
-The [`randomize_lt()`](/languages/func/stdlib#randomize-lt) function mixes the transaction's logical time into the random seed before generating random numbers. This provides basic entropy from blockchain state.
+<Aside type="tip">
+  When choosing an approach, consider the value at risk and required time-to-finality. For high-stakes applications such as lotteries with significant funds, use commit-reveal. Audit implementations through formal verification when possible.
+</Aside>
+
+## Single-block randomness
+
+The [`random.initialize()`](/tolk/features/standard-library#random-initialize) function initializes the random number generator with a seed derived from the transaction's logical time. This provides basic entropy from blockchain state.
```
delovoyhomie: `random.initialize()` maps to the `randomize_lt` behavior: it mixes the transaction logical time into the current per-contract random seed

to me this sentence currently sounds like the seed is derived from logical time alone, which conflicts with the later block-seed/address explanation and can understate the validator-controlled seed component

I'd consider wording this as "mixes the transaction's logical time into the current random seed" instead of "initializes ... with a seed derived from..."

contract-dev/techniques/random.mdx
```diff
@@ -5,164 +5,142 @@ sidebarTitle: "Random numbers"
...
+- Systems requiring Byzantine fault tolerance.
 
 ## Best practices
 
-- Avoid standalone [`random()`](/languages/func/stdlib#random) calls. Validators controlling the block seed can predict the output.
-- Keep randomness out of external message receivers. External messages remain vulnerable even with [`randomize_lt()`](/languages/func/stdlib#randomize-lt).
-- Use hybrid or off-chain entropy for critical applications. Combine on-chain randomness with off-chain entropy or external randomness oracles when significant value is at risk.
-- Test randomness behavior across different blocks. Verify that the contract behaves correctly when randomness is manipulated within validator capabilities.
+- Always call `random.initialize()` before using `random.uint256()` or `random.range()` this prevents users from predicting values but does not protect against validators who control the block seed.
+- Keep randomness out of external message receivers. External messages remain vulnerable even with `random.initialize()`.
```
delovoyhomie: doesn't explain why external messages are uniquely worse. the actual reason: external messages don't transfer funds, so an attacker can resubmit them for free until they get a favorable random outcome. internal messages cost gas to send, which raises the cost of retry attacks. worth a sentence:

```suggestion
+Keep randomness out of external message receivers. Because external messages do not transfer funds, an attacker can resubmit them for free until the random outcome is favorable - so `random.initialize()` does not protect external entry points.
```

contract-dev/techniques/random.mdx
```diff
@@ -5,164 +5,142 @@ sidebarTitle: "Random numbers"
...
-Call [`randomize_lt()`](/languages/func/stdlib#randomize-lt) once before using [`random()`](/languages/func/stdlib#random) or [`rand()`](/languages/func/stdlib#rand) functions. The logical time adds variability to the seed, making it harder for observers to predict the outcome without executing the transaction.
+Call `random.initialize()` once before using [`random.uint256()`](/tolk/features/standard-library#random-uint256) or [`random.range()`](/tolk/features/standard-library#random-range-limit) functions. The logical time adds variability to the seed, making it harder for observers to predict the outcome without executing the transaction.
 
-```func
-randomize_lt();
+```tolk title="Tolk"
+random.initialize();
 // Generates values from 0 to 99
-int random_number = rand(100);
+var randomNumber = random.range(100);
```
delovoyhomie: randomNumber is never reassigned. idiomatic Tolk uses `val` for immutable bindings:

```suggestion
+val randomNumber = random.range(100);
```

delovoyhomie (dismissed):

ijonele (approved):

PR #2131, opened 2026-04-24
feat(sharding): put in agreement with styleguide and port code samples to Tolk

### LINKED ISSUE #500: [Contract dev > Contract sharding] Add more examples and code with different scale & scope
Currently, there's a bleak example in Tact, but it'd be better to show more things of increasing difficulty and interesting ideas: starting from the most basic counter/ping-pong contract and ending with Jettons and NFTs.

That is, examples should cover:
- something simple and non-sharded
- most basic `Parent` / `Child` contracts and interaction of
- [open-ended poll contracts](https://github.com/tact-lang/create-tact/pull/9/files#diff-678de6e5bfe1ab9cfa8e565bdd4b7b8d93234d53347b7d100daf561da4c1cf77) (Poll and Voter) ported to Tolk — it's interesting because it makes active use of bounced messages in its sharding logic
- finally, either show smaller snippets of or just directly link to Jetton and NFT implementations and/or docs

### LINKED ISSUE #1440: [Contract dev > Contract sharding] Change code samples to use Tolk
Part of #727

### LINKED ISSUE #2106: [Contract dev > Contract sharding] Put in agreement with styleguide
- Put in agreement with styleguide
- Convert example from Tact to Tolk

aigerimu:
Closes https://github.com/ton-org/docs/issues/1440
Closes https://github.com/ton-org/docs/issues/2106

## Summary by CodeRabbit

* **Documentation**
  * Expanded contract sharding guidance with clearer TON state limit explanations and updated NFT and jetton examples to follow the refined parent/child model.
  * Added a dedicated section on deterministic child contract address derivation from `StateInit`, including recommended message verification from the parent.
  * Refreshed ton-decompiler CLI wording for clarity (“Install the tool suite:”).
* **Chores**
  * Updated formatter configuration to explicitly enable bracket spacing.


contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,153 @@ title: "Contract sharding"
...
+
+// Parent deploys another child when it receives this message.
+struct (0x5b6f1392) DeployAnother {}
+
+// Build StateInit for a TodoChild instance with the given seqno.
+fun calcDeployedTodoChild(
+    seqno: uint64,
+    todoChildCode: cell,
+): AutoDeployAddress {
+    val childStorage: TodoChildStorage = {
```
delovoyhomie: I guess the child `StateInit` only includes `seqno`, so two parents using the same child code and sequence number can derive the same child address. it also contradicts the caution above, because the child has no stored parent address to verify 

 to fix it is to include the parent address in the child's initial storage, pass it into address derivation, and verify it in the child handler

content/contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,153 @@ title: "Contract sharding"
...
+```tolk title="parent.tolk"
+import "storage"
+
+type TodoParentMessage = DeployAnother
+
+fun onInternalMessage(in: InMessage) {
+    val msg = lazy TodoParentMessage.fromSlice(in.body);
+
+    match (msg) {
+        DeployAnother => {
```
delovoyhomie, line 154: `DeployAnother` is accepted from any sender and spends `ton("0.1")` from the parent. If copied as-is, anyone can repeatedly trigger child deployments and drain the parent balance

either mark this handler as demo-only, or add authorization. For example:

```diff
 struct TodoParentStorage {
+    // Initialize this field in the parent's StateInit during deployment.
+    adminAddress: address
     numChildren: uint64 = 0
     // Parent must know the child contract code to deploy new instances.
     todoChildCode: cell
 }
```

```diff
         DeployAnother => {
             var storage = lazy TodoParentStorage.load();
+            assert (in.senderAddress == storage.adminAddress) throw 0xFFFF;
             storage.numChildren += 1;
```

content/contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,153 @@ title: "Contract sharding"
...
+            var storage = lazy TodoParentStorage.load();
+            storage.numChildren += 1;
+
+            // Send a message to the auto-calculated address and attach
+            // the child code and initial data so the child is deployed.
+            val deployMsg = createMessage({
```
delovoyhomie, line 163: the `createMessage` call omits explicit bounce behavior. The Tolk message docs say `createMessage` should specify bounce behavior, and the full examples in this repo do so 

https://docs.ton.org/languages/tolk/features/message-handling#bouncemode-in-createmessage

I'd add an explicit mode, for example:

```diff
             val deployMsg = createMessage({
+                bounce: BounceMode.Only256BitsOfBody,
                 dest: calcDeployedTodoChild(
                     storage.numChildren,
                     storage.todoChildCode,
                 ),
```

contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,153 @@ title: "Contract sharding"
 import { Aside } from '/snippets/aside.jsx';
 import { Image } from '/snippets/image.jsx';
 
-Some protocols need to store a lot of information in contracts, for example, tokens that have many users. In TON, there is a limit on how much can be stored in a single contract. The solution in TON is to split the data across many different contracts, where you can quickly find the right contract by a key and retrieve the required information from it.
+Some protocols need to store a lot of information in contracts, for example, token contracts with many users. In TON, there is a limit on how much can be stored in a single contract. The solution is to split the data across multiple contracts.
 
-In such protocols, there is a child contract that initially contains the information identified by a key. In some protocols, it is important to know the Parent contract, which acts as the information manager.
+Each such contract, referred to as a child contract, is associated with a key that allows direct access to the required contract and its data. Some protocols also introduce a _parent_ contract that coordinates child contracts.
```
delovoyhomie: "a key that allows direct access to the required contract and its data" repeats "contract" a bit awkwardly. it makes sense to read cleaner as something like: `a key that uniquely determines its address`

content/contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,153 @@ title: "Contract sharding"
...
+import "storage"
 
-    numChildren: Int as uint64;
+type TodoChildMessage = Identify
 
-    init() {
-        self.numChildren = 0;
-    }
+fun onInternalMessage(in: InMessage) {
+    val msg = lazy TodoChildMessage.fromSlice(in.body);
```
delovoyhomie, line 98: I'm not sure, but I think the `else` branch in the `match` below may not be reachable as written. Since `TodoChildMessage = Identify` is a single-variant union, `lazy TodoChildMessage.fromSlice(in.body)` should throw on any body that doesn't parse as `Identify` - including an empty body, since the opcode parse fails first

If the intent is to ignore empty top-up messages, the empty-body check probably needs to happen before the `lazy fromSlice` call for instance

```diff
 fun onInternalMessage(in: InMessage) {
+    if (in.body.isEmpty()) {
+        return;
+    }
     val msg = lazy TodoChildMessage.fromSlice(in.body);
```

it definitely worths double-checking against actual Tolk semantics

content/contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,153 @@ title: "Contract sharding"
...
+
+type TodoParentMessage = DeployAnother
+
+fun onInternalMessage(in: InMessage) {
+    val msg = lazy TodoParentMessage.fromSlice(in.body);
+
+    match (msg) {
+        DeployAnother => {
+            var storage = lazy TodoParentStorage.load();
+            storage.numChildren += 1;
```
delovoyhomie, line 159: minor point: `numChildren` increments before the send, and the send uses `SEND_MODE_IGNORE_ERRORS`. If the action fails to queue, the counter still bumps, so the next deploy uses `seqno = N+2` and leaves a gap

absolutely not a bug if contiguous numbering doesn't matter, but might be worth a sentence in the prose or a short comment in the sample so readers don't assume sequence numbers are guaranteed dense :)

contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,157 @@ title: "Contract sharding"
...
 >
-  Child contracts should store information about the Parent so that only it can authorize important state changes.
+  Child contracts should store the parent contract address and verify that incoming messages originate from it. This ensures that only the parent contract can perform authorized state changes.
 </Aside>
 
+## NFT and jetton examples
+
+Consider NFTs: the collection acts as the parent contract, and each NFT item is a child contract. The key in this case is the item index, and only the collection can set the initial owner.
+
+For jettons, the parent contract is the minter, and the child contracts are user wallets. The key is the user's smart contract address, and the value is the user's token balance.
```
ArkadiyStena:
```suggestion
For jettons, the parent contract is the minter, and the child contracts are user wallets. The key is the user's smart contract address.
```

contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,157 @@ title: "Contract sharding"
...
+
+Both patterns follow the same principle: each key maps to a separate contract. In jetton protocols, there is a unique contract per user, while in NFT collections, there is one contract per item (by index) that is shared across all users.
```
ArkadiyStena: > that is shared across all users.

This wording sounds a bit strange.

content/contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,157 @@ title: "Contract sharding"
...
+}
 
-In general, infinite data structures that can actually scale to billions are very difficult to implement on blockchain efficiently. This pattern showcases the power of TON.
+fun TodoChildStorage.load() {
+    return TodoChildStorage.fromCell(contract.getData())
+}
 
-```tact Tact
-import "@stdlib/deploy";
+fun TodoChildStorage.save(self) {
```
ArkadiyStena, line 58: we should add `parentAddress: address` here

contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,157 @@ title: "Contract sharding"
...
-    receive("identify") {
-        dump(self.seqno);
+// Child prints its sequence number when it receives this message.
+struct (0x49f29a21) Identify {}
+
+// Parent deploys another child when it receives this message.
+struct (0x5b6f1392) DeployAnother {}
+
+// Build StateInit for a TodoChild instance with the given seqno.
+fun calcDeployedTodoChild(
```
ArkadiyStena: maybe we should move it to `parent.tolk` (I am not sure)

ArkadiyStena (dismissed):

contract-dev/techniques/contract-sharding.mdx
aigerimu: Addressed by moving `calcDeployedTodoChild` to `parent.tolk`. But this still needs to be validated

cc @Shvandre @Kaladin13

content/contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,168 @@ title: "Contract sharding"
...
 >
-  Child contracts should store information about the Parent so that only it can authorize important state changes.
+  Child contracts should store the parent contract address and verify that incoming messages originate from it. This ensures that only the parent contract can perform authorized state changes.
 </Aside>
 
+## NFT and jetton examples
+
+Consider NFTs: the collection acts as the parent contract, and each NFT item is a child contract. The key in this case is the item index, and only the collection can set the initial owner.
+
+For jettons, the parent contract is the minter, and the child contracts are user wallets. The key is the wallet's smart contract address.
```
delovoyhomie: small accuracy thing: "The key is the wallet's smart contract address" is ambiguous - and probably wrong as it reads.the previous sentence says the children are user wallets, so "the wallet's smart contract address" naturally refers to the child's own address, which is the output of address derivation, not the input/key

the key for a jetton wallet is the address of the wallet's owner (the user's main wallet contract) which gets fed into address derivation along with the minter address and the wallet code

```diff
-For jettons, the parent contract is the minter, and the child contracts are user wallets. The key is the wallet's smart contract address.
+For jettons, the parent contract is the minter, and the child contracts are jetton wallets. The key is the address of the jetton wallet's owner, and the value is the owner's jetton balance.
```

content/contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,168 @@ title: "Contract sharding"
...
-        self.numChildren = 0;
+type TodoChildMessage = Identify
+
+fun onInternalMessage(in: InMessage) {
+    val msg = lazy TodoChildMessage.fromSlice(in.body);
+
+    match (msg) {
+        Identify => {
+            val storage = lazy TodoChildStorage.load();
+            assert (in.senderAddress == storage.parentAddress) throw 0xFFFF;
```
delovoyhomie: `throw 0xFFFF` is reused for three different conditions: "sender is not the parent" (here), "unexpected message body" ([line 101](https://github.com/ton-org/docs/blob/contract-sharding-polishing/contract-dev/techniques/contract-sharding.mdx#L101)), and "sender is not the admin" ([line 144](https://github.com/ton-org/docs/blob/contract-sharding-polishing/contract-dev/techniques/contract-sharding.mdx#L144)). same code for three distinct failure modes makes a tx-dump in an explorer harder to triage

worth picking distinct codes (any value in 32-65535 outside reserved ranges), e.g. `100` / `101` / `200`, but I am not sure

content/contract-dev/techniques/contract-sharding.mdx
```diff
@@ -5,80 +5,168 @@ title: "Contract sharding"
...
+## Unbounded data structures
 
-For Jettons, the Parent is the minter and the Children are user wallets. The key is the user's smart contract address, and the value is the user's token balance.
+Contract sharding supports an unbounded number of potential child contracts.
 
-In general, jettons and NFTs share this principle, but broadly speaking, jetton protocols have a unique contract per user, while NFTs have a single contract per item (by index) that is shared across all users.
+In general, data structures that can scale to very large sizes are difficult to implement efficiently on blockchains. This pattern allows such scaling by distributing data across multiple contracts.
 
-## Unbounded data structures
+The following example shows a parent contract that deploys child contracts and assigns each child a sequence number. The shared storage file defines the data layouts and message types used by both contracts.
```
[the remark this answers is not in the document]
novusnota: Good catch, swapped issues.

novusnota (approved):

PR #2171, opened 2026-05-18
feature: fumadocs

Alejandbel:


novusnota:
Thanks for the initial setup — it was a useful reference when moving to Fumadocs. Now, the documentation uses the framework in (almost) full capacity, see https://ton-org.github.io/docs. Some things, such as OG image generation and LLM-specific OpenAPI page generation, will be addressed shortly.

The diverged content is present and up-to-date in https://github.com/the-ton-tech/ton-docs-private — I'll be looking into it soon, especially when migrating the contents of TON Connect, AppKit, and WalletKit sections.

As such, this PR will be closed.

Issue #2192, opened 2026-06-02
Migration to Fumadocs


novusnota:
The rough plan:

- Gradually move all pages
  - [x] Add required components
    - [x] ~~Port Aside -> replace with Callouts or have a mapping~~
    - [x] ~~Port FileTree -> map onto existing components~~
    - [x] Port Image -> combine with img or ImageZoom
    - [x] Port Cards: remove the outer border for icons, use gradient
    - [x] Port other Mintlify components in use (`<[A-Z]`)
      - [x] ~~Create stubs for all previously used ones~~
    - (this can wait) Edit `contribute/` pages and overviews
    - [x] Adjust CodeGroups and enhance code blocks:
      - (this can wait) Requires an `Icon` port: this might (?) include fontawesome icons set too, as an alternative
      - [x] ~~`title` alias for `tab`~~
      - (this can wait) an `icon` prop for tabs: use `<Icon>` component to add `icon` in tab and code block components via a Shiki or remark transformation, not the augmentation of the underlying components. Yet, icon handling in CodeBlocks has to be modified slightly with `icon=""` present. Also, notice the semantical difference of `title` and `tab` meta attrs.
        - https://github.com/fuma-nama/fumadocs/blob/dev/packages/core/src/mdx-plugins/rehype-code.core.ts#L48
        - https://github.com/fuma-nama/fumadocs/blob/dev/packages/core/src/mdx-plugins/transformer-icon.ts#L132 (override with a custom transformer that uses `Icon` component or just lucide icons directly)
        - https://github.com/fuma-nama/fumadocs/blob/dev/packages/core/src/mdx-plugins/remark-code-tab.ts
        - https://github.com/fuma-nama/fumadocs/blob/dev/apps/docs/source.config.ts
    - [x] Fix issues with styles on selected pages, especially with per-page components: TVM instructions, Catchain visualizer
  - [x] ~~Frontmatter updates~~
    - [x] ~~`wide` pages → switched to using the built-in `full` prop~~
    - [x] ~~`openapi` pages: contents, generation, sidebar placement~~
    - [x] ~~`noindex` support~~
    - [x] ~~`url` support~~
    - [x] ~~`sidebarTitle` support~~
    - [x] ~~`tag` support~~
  - [x] ~~Use Inter font~~
  - [x] ~~Use brand colors, use TON logo~~
  - [x] ~~`meta.json` with previous structure~~
- [x] ~~Gradually move prior top-level setup~~
[...]

PR #2207, opened 2026-06-04
fix: root-relative resource link postprocessing

novusnota:
Towards #2192. An attempt, which might not be successful. In that case, let's process the `out/` folder after the build.

## Summary by CodeRabbit

**Bug Fixes**
* Improved handling of relative media and link attributes in markdown content to ensure consistent path rewriting across different element types and components.


novusnota:
Ok, that did not work, so let's just try to process the resulting output in `out/` folder then.

PR #2257, opened 2026-07-09
feat: add Tonviewer to explorers list

klpx:


novusnota:
@klpx thanks, but we're currently reducing the number of referenced commercial products in the docs. The present situation is already a compromise :)

Sorry for the inconvenience :pray: 

That said, an open-source, dev-first explorer _might_ be released in the near future. If or when that happens, it will definitely be featured here and across the docs.

PR #2290, opened 2026-08-10
feat(api): create WebSocket AsyncAPI spec for TON Center's Streaming API

novusnota:
An experiment.


ebijoo (commented):
Hay

novusnota:
It does not bring any value on top of existing documentation, and the proper AsyncAPI spec as generated here should probably be hosted off of TON Center's Streaming API part of API v3 instead. I might be wrong, so the branch will not be deleted as to have an opportunity to return to this in the future.

Issue #34, opened 2025-09-03
LLM checking CI


verytactical:
Add CI step that runs an LLM that checks

- changed files follow the styleguide
- grammar is correct
- changes are consistent with the rest of documentation, i.e. no changes were forgotten

Issue #23, opened 2025-09-03
Internal documents on organizational principles of the structure


novusnota:
"Where to put X?"

PR #2309, opened 2026-08-28
feat(docs): add heading aliases

Danil42Russia:
Now, for anchors such as `\d-and-\d-and-\d`, alternatives will be automatically added for each `\d`.

For example: `12-and-23-and-34` will result in the anchors `#12-and-23-and-34`, `#12`, `#23` and `#34`.

### DIFF
--- a/source.config.ts
+++ b/source.config.ts
@@ -208,6 +208,21 @@ export default defineConfig({
       // Default Fumadocs remark plugins
       ...v,
       // Additional plugins
+      function remarkHeadingAliases() {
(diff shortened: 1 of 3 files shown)


novusnota (approved):
Original solution was interesting, but it had many potential issues that could've bit us later on. Besides, we'll only need such anchor modifications on very few selected pages. As such, I've remade the solution into a rehype (html-level) plugin as opposed to the original remark (markdown-level) one.

Thanks, will be merging this shortly!

